PR #214, opened 2014-04-14
OptiFine support, second try

02JanDal:


peterix:
This is obsolete. Wonko repo will replace it cleanly (or at least, one would hope).

PR #363, opened 2014-07-13
Automatically copy added mods to the central mods folder

02JanDal:


peterix:
I'm not sure if I really want this in mmc. I'm leaning towards 'no', therefore denied.

PR #550, opened 2014-10-15
Fixed build on OS X 10.10

miuirussia:


peterix:
Well, it doesn't break on our current build server and I would rather keep it that way without experiments.

PR #592, opened 2014-11-09
Fix #209: Allow overriding the default account per instance

### LINKED ISSUE #209: Enable setting default account for Instances
Would be a nice feature aswell.

peterix:
Let's see what this is about.


peterix:
I'd say will need a complete re-do with the new account system from multilaunch and would only cause a rebase cascade ...

PR #607, opened 2014-11-14
Add a list view mode

02JanDal:


peterix:
The tree view should start expanded.

peterix:
The proxy model has some bugs in it.

I ended up with this after using the settings:
badness1

Many of the groups are suddenly empty.

peterix:
Please fix and make a new pull request. This is just too buggy.

PR #638, opened 2014-11-28
Drag & drop mods into instances

02JanDal:


peterix:
This is a cool concept, but with so many minecraft versions involved, and no way to filter by version, ultimately not useful.
At the same time, I can't drag in mods from an external exporer-like window, where I have mods cataloged by MC version and I'm sure what's happening.

It also takes up main window real estate.

peterix:
Rebase causes conflicts.

peterix:
whatwhere

Now, what do I put where? I honestly have no idea. Some of the mods, if they are mods, have no version. Some of the components are random text files.

PR #707, opened 2015-01-03
Update and sort modlist after adding mods

robotbrain:


peterix:
Needs to be tested with legacy jar mods first.

PR #765, opened 2015-02-02
Change default PermGen from 64 to 128 mb

Greenphlem:
This will help users who run mods. Too many issues have been made over the lack of more permgen being assigned.

### DIFF
--- a/MultiMC.cpp
+++ b/MultiMC.cpp
@@ -511,7 +511,7 @@ void MultiMC::initGlobalSettings(bool test_mode)
 	// Memory
 	m_settings->registerSetting({"MinMemAlloc", "MinMemoryAlloc"}, 512);
 	m_settings->registerSetting({"MaxMemAlloc", "MaxMemoryAlloc"}, 1024);
-	m_settings->registerSetting("PermGen", 64);
+	m_settings->registerSetting("PermGen", 128);
 
 	// Java Settings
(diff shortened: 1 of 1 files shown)


Heufneutje:
Won't this change cause the warning to appear for everyone who uses Java 8 though? There's a lot of people complaining about that too.

Kilobyte22:
Does MultiMC even set permgen when java8 is in use?

02JanDal:
Yes, since we currently don't have a reliable way of determining which version of Java is used.

Kilobyte22:
Wouldn't a java --version do the job?

02JanDal:
Nope, the output isn't really reliable. We currently have a simple java program that dumps the java version from a system property, but going from there to the boolean value of (java < 8) is the hard part.

peterix:
It can be done at the cost of running more java processes. Run java once to find out what version you are working with, run it second time for real, with the right parameters.

Heufneutje:
And besides, once Java 7 support is dropped the problem should sort itself out.

Greenphlem:
Right but currently more than around 50% of purple use java 7 or less, so
it's still a problem.

Greenphlem:
People* idiot phone

peterix:
Well, merged. And added the necessary bits for the warnings from java 8 to not show up in the console.

PR #792, opened 2015-02-19
Clean up/Qt-ify the updater

### LINKED ISSUE #77: Update Problem with bad windows Username - minor bug
If your windows username contains characters like ášč atd. then it crashes with 
file does not exist.
Uploading  . . .

02JanDal:
Should fix #77 and other updater-related annoyances.

* Replace anyoption by QCommandLineParser
* Use Qt for filesystem access
* Replace tinyxml by QXmlStreamReader
* Switch non-gui parts to QString
* Fix various bugs
* Format files
* Add an integration test
* Remove unused files
* Remove relaunching and elevated mode

Tested on Linux and OS X, still needs testing on windows.

This will require a static Qt to be available on the builders (QtCore only).


02JanDal:
Mainly missing a static Qt build on all the builders. I can build and package them for linux and osx but not for windows.

Also needs to be tested on windows.

peterix:
Decided to not use the updater at all. Using the branch only for reference and as a source of useful code bits.

PR #996, opened 2015-05-22
NOISSUE Add cmake format target

robotbrain:
Adds a make target that formats all source code using clang-format
Currently separate targets for each component, logic and application

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,6 +131,8 @@ include(Coverity)
 
 ############################### Built Artifacts ###############################
 
+include(Format)
+
 add_subdirectory(tests)
 
 add_subdirectory(logic)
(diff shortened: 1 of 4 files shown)


peterix:
This requires #1037

peterix:
Rebased it, tried it out.

I'm not sure if I want to reformat everything blindly.

Example:
```
diff --git a/tests/data/1.json b/tests/data/1.json
index 0f2d6ce..48c32e7 100644
Binary files a/tests/data/1.json and b/tests/data/1.json differ
```

A lot of the manual formatting is gone after this is used - obviously. And that's not nice.

Plus it adds a lot of unrelated changes that break branches for no good reason.

Maybe we need a better clang-format config that produces something closer to the existing code.

peterix:
I won't do this.

PR #1012, opened 2015-05-28
Various changes from multiauth that are unrelated to it

02JanDal:
Since the multiauth PR got quite big I split of all changes that weren't actually part of the auth things.

This PR does not introduce any big behavioral differences, and contains things that would be nice to have even if multiauth doesn't end up getting merged.


peterix:
Merged, found a lot of weird things, fixed some of it.

Reverted, for now.

PR #1042, opened 2015-06-04
Various changes from multiauth that are unrelated to it

peterix:
Another attempt.


application/MultiMC.cpp
```diff
@@ -331,6 +333,45 @@ void MultiMC::initIcons()
...
+			else
+			{
+				return QPixmap();
+			}
+		});
+	Resource::registerTransformer<QByteArray, QPixmap>([](const QByteArray &data) -> QPixmap
+													   {
+														   return QPixmap::fromImage(
+															   QImage::fromData(data));
+													   });
```
02JanDal: Formatting derp?

logic/resources/Resource.h
```diff
@@ -0,0 +1,99 @@
...
+	Ptr placeholder(Ptr other);
+
+	/// Use these functions to specify what should happen when e.g. the resource changes
+	Ptr applyTo(ResourceObserver *observer);
+	Ptr applyTo(QObject *target, const char *property = nullptr);
+	template<typename T, typename Func>
+	Ptr then(Func func)
+	{
+		return applyTo(new FunctionResourceObserver<T,Func>(func));
+	}
```
02JanDal: Ok, it's not as bad as I thought. Though after you found the std::function<Func>::{argument_type,result_type} I think it still would make sense to add automatic detection back.
Return type always being void: alright for now, but that means you can't call then(&SomeClass::someMember); if someMember has a non-void return value, you'd have to wrap it in a lambda.

02JanDal:
Alright, after what you said on IRC I had expected way worse, this is still ok, though I still would like to add back type detection for then() with std::function at some point.

logic/resources/Resource.h
peterix: Well, the issue I ran into is this:
* I need to convert a lambda to a std::function.
* std::function, when declared as a class member, can't quite use auto.
* Therefore I need to know the return type and argument(s) of the lambda
* There's no sane way that I could come up with to do that without a pile of trait structs, just like you had.

Therefore, this.

I learned more about this side of c++ than I did in the last 10 years... and I still don't like it.

application/MultiMC.cpp
peterix: Guess so.

PR #1094, opened 2015-06-26
Set up build system for Squish Coco coverage reporting

02JanDal:
This PR contains two separate commits. The first adds [Squish Coco](http://www.froglogic.com/squish/coco/) coverage reporting support to the build system. To use this, run (in a different build directory)
`cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/SquishCocoGCC.cmake ..`.
You will then get access to targets like `coverage`, `coverage_browse` and
`coverage_report_html`.

* **coverage** will create a Megatron.csmes file which can be used for further processing
* **coverage_browse** will open `coveragebrowser` with the created csmes file
* **coverage_report_html** will generate HTML documentation in `coverage/reports/html_html`

You can also set `SQUISHCOCO_HTML_PUBLISH` to `ON` to get access to the
`coverage_publish_html` target. It is configured using the variables
`SQUISHCOCO_HTML_PUBLISH_HOST`, `SQUISHCOCO_HTML_PUBLISH_USER` and
`SQUISHCOCO_HTML_PUBLISH_DIR`. It will use the information from these
to upload the HTML coverage report to a remote server using rsync.

The second commit adds unit tests for AbstractCommonModel, including fixes for some bugs that were discovered during the creation of the tests.


peterix:
I am simply not going to do this. The thing costs 3000 euro just to get it. And then 1000 euro every year.

Simply not an option.

Issue #1047, opened 2015-06-05
Add save/world management


ThePixelNerd:
Ok, first I want to say this application is absolutely invaluable around my house! With 4 of us that play Minecraft, it's a lifesaver being able to use this program!

Ok, my question is, where can I find the skin that I see on the screenshots? It's where the icons look like grass blocks....I don't see the option for that skin anywhere :(

My request is, in the context menu for each instance of the game it shows such things as: 
Edit Instance
Instance Settings
Edit Notes
Manage Screenshots
Etc.
I would like to see the option of: "Open Map/Save Folder" or however you would want to word it.

My kids are constantly downloading maps and it's a headache to tell them or anyone else I might be teaching how to use MultiMC to go into instance folder, minecraft, find the saves folder, extract the save into that folder blah blah blah.

It would be awesome to have a one click link that opened the Saves Folder....or if there was some kind of option to just upload a new save straight into there through the program?!?!?

Thanks again for this amazing program that is miles above all others! God Bless!

Kilobyte22:
Maybe a map management GUI? Wasn't there something like that already? Or was that MultiMC 3/4?

peterix:
It was. And will fit into MMC5 much better.

Actually, I'm just going to do this, because it sounds like fun.

peterix:
There are icon themes in the settings:
Settings -> User Interface -> Icon Theme.

UI themes in general aren't there yet - whatever the OS can offer is used.

peterix:
Should be just like the other folder management pages, except:
- [ ] Add MCEdit integration
- [ ] Add export, not just import
- [ ] Add import from URL
- [ ] Add size column

PR #1220, opened 2015-08-30
GH-1047 World management for instances. Removal only currently.

robotbrain:


application/pages/WorldListPage.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <QKeyEvent>
+
+WorldListPage::WorldListPage(BaseInstance *inst, std::shared_ptr<WorldList> worlds, QString id,
+							 QString iconName, QString displayName, QString helpPage,
+							 QWidget *parent)
+	: QWidget(parent), ui(new Ui::WorldListPage)
+{
+	ui->setupUi(this);
+	ui->tabWidget->tabBar()->hide();
+	m_inst = inst;
```
02JanDal: Might want to put all of these into the initialization list of the constructor instead.

application/pages/WorldListPage.cpp
```diff
@@ -0,0 +1,94 @@
...
+	m_worlds = worlds;
+	m_id = id;
+	m_displayName = displayName;
+	m_iconName = iconName;
+	m_helpName = helpPage;
+	ui->worldTreeView->setModel(m_worlds.get());
+	ui->worldTreeView->installEventFilter(this);
+	auto smodel = ui->worldTreeView->selectionModel();
+	connect(smodel, SIGNAL(currentChanged(QModelIndex, QModelIndex)),
```
02JanDal: Prefer "new" connect syntax (function pointers)

application/pages/WorldListPage.cpp
```diff
@@ -0,0 +1,94 @@
...
+			SLOT(modCurrent(QModelIndex, QModelIndex)));
+}
+
+void WorldListPage::opened()
```
02JanDal: If I remember that code correctly there is a lot of duplication between this the page for mods. Maybe extract into common base class (BaseFolderPage or similar)?

application/pages/WorldListPage.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+bool WorldListPage::shouldDisplay() const
+{
+	if (m_inst)
+		return !m_inst->isRunning();
+	return true;
+}
+
+bool WorldListPage::worldListFilter(QKeyEvent *keyEvent)
```
02JanDal: worldListFilter and eventFilter could be heavily minimized (3-4 lines), if you don't plan to add more stuff to them.

application/pages/WorldListPage.h
```diff
@@ -0,0 +1,69 @@
+//
+// Created by robotbrain on 8/18/15.
```
02JanDal: Proper license headers (same for other new files).

logic/minecraft/LegacyInstance.cpp
```diff
@@ -319,6 +320,16 @@ std::shared_ptr<ModList> LegacyInstance::texturePackList() const
 	return texture_pack_list;
 }
 
+std::shared_ptr<WorldList> LegacyInstance::worldList() const
```
02JanDal: OneSixInstance::worldList and this worldList are basically identical, so the function should probably be moved up in the inheritance hierarchy.

logic/minecraft/World.h
```diff
@@ -0,0 +1,47 @@
...
+	World(const QFileInfo &file);
+	QString name() const
+	{
+		return m_name;
+	}
+	bool isValid() const
+	{
+		return is_valid;
+	}
+	// delete all the files of this mod
```
02JanDal: Copy/Paste miss: still using "mod"

logic/minecraft/WorldList.cpp
```diff
@@ -0,0 +1,226 @@
...
+WorldList::WorldList(const QString &dir)
+	: QAbstractListModel(), m_dir(dir)
+{
+	ensureFolderPathExists(m_dir.absolutePath());
+	m_dir.setFilter(QDir::Readable | QDir::NoDotAndDotDot | QDir::Files | QDir::Dirs |
+					QDir::NoSymLinks);
+	m_dir.setSorting(QDir::Name | QDir::IgnoreCase | QDir::LocaleAware);
+	m_watcher = new QFileSystemWatcher(this);
+	is_watching = false;
+	connect(m_watcher, SIGNAL(directoryChanged(QString)), this,
```
02JanDal: Signal slot syntax again (won't comment if it's in more places after this)

application/pages/WorldListPage.cpp
robotbrain: I don't really want to do heavy refactoring for a simple feature at this point

application/pages/WorldListPage.cpp
robotbrain: adding (import) might be added, so leaving for now

logic/minecraft/WorldList.h
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright 2015 <copyright holder> <email>
```
02JanDal: Probably should look over all license headers, I saw 2013-2015 being used in one new file too.

logic/minecraft/WorldList.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+QVariant WorldList::data(const QModelIndex &index, int role) const
+{
+	if (!index.isValid())
+		return QVariant();
+
+	int row = index.row();
+	int column = index.column();
+
+	if (row < 0 || row >= worlds.size())
```
02JanDal: This and other places: use brackets even for single-line bodies.

logic/minecraft/WorldList.h
```diff
@@ -0,0 +1,101 @@
...
+{
+    Q_OBJECT
+public:
+    enum Columns {
+        NameColumn
+    };
+
+    WorldList ( const QString &dir );
+
+    virtual QVariant data ( const QModelIndex &index, int role = Qt::DisplayRole ) const;
```
02JanDal: This entire file could use some clang-formatting (spaces everywhere, newlines before {, ; after function body and more)

application/pages/WorldListPage.cpp
02JanDal: Ok, just watch out, there are tons of copy/paste issues everywhere in this PR...

application/pages/WorldListPage.cpp
02JanDal: With just a few more things I'd still change these. You can't tell at a quick glance what these functions are supposed to do (which should be possible since what they're doing is quite trivial).

PR #1225, opened 2015-09-05
Fixed overwriting of wrapper command

iambob314:
Fixed "WrapperCommand" setting being overwritten by the value of "PreLaunchCommand".


peterix:
Alright. That was a silly copypasta oversight :)

iambob314:
Thanks!

PR #1259, opened 2015-09-27
Mod download using a built-in browser

02JanDal:
When a URL is dropped on any ModFolderPage (mods, coremods, texturepacks, resourcepacks) a built-in browser is opened, allowing the user to navigate through adfly etc. Once the actual download starts it will automatically be put in the correct place.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,7 @@ os:
  - linux
 # - osx
 env:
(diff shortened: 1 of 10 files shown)


peterix:
Hm. We might be able to recycle this. For now, I'll just link it from the Wonko part of workflowy.

PR #1308, opened 2015-10-31
Make translations respect data path, on Linux.

Spencer-H:
This allows the path where translations are installed to also be controlled by the -d argument. There is probably a better way to handle this, but it isn't good form to require write access to the directory above a binary on Linux. This also just makes MultiMC confine its files to one directory. Which is very useful if you launch it with the -d ~/.multimc option.

### DIFF
--- a/application/MultiMC.cpp
+++ b/application/MultiMC.cpp
@@ -172,8 +172,7 @@ MultiMC::MultiMC(int &argc, char **argv, bool test_mode) : QApplication(argc, ar
 
 // static data paths... mostly just for translations
 #ifdef Q_OS_LINUX
-	QDir foo(FS::PathCombine(binPath, ".."));
-	staticDataPath = foo.absolutePath();
+	staticDataPath = dataPath;
 #elif defined(Q_OS_WIN32)
 	staticDataPath = binPath;
 #elif defined(Q_OS_MAC)


02JanDal:
Looks good and works in my development environment (haven't tested as installed, but it looks like it should work).

@peterix will need to approve.

Spencer-H:
Thanks. Yeah I'm running MultiMC with this on my system and it's quite nice to be able to put MultiMC and jars in /usr/share/multimc and have it put everything it needs to run in ~/.multimc

02JanDal:
Note that putting MultiMC in a non-writable location is not officially supported, and you may run into other issues down the road.

Spencer-H:
MultiMC doesn't seem as if it's incapable of running in a non-writeable location, it just disabled the auto-update, which the packaging system would handle anyway. I'll deal with any issues that do come up though, and offer them as pull requests if my solutions seem sane enough.

peterix:
I wouldn't call it incapable. It's just not the way it is built and packaged by default. So I won't notice if it breaks.

02JanDal:
The issue is that we currently have to support several external systems (Mojang, Forge etc.), and they can (and have in the past) break things for us, in which case we want, and need, to bring out updates fast. Thus we use our own updater, since distro packages are usually way to far behind. Therefore installations not using our updater will not be supported. (we are working on moving away from directly depending on those external systems though, once that's done installations without our updater may become more viable)

Secondly, as peterix said, this is not the way the program is packaged and tested by default, thus we wouldn't notice if it breaks. This means a significant increase in maintenance cost.

I still think this PR is good for other reasons, and should be merged though.

Spencer-H:
Very few packages on Void seem to be officially supported anyway. If something breaks on Void Linux just point them my way. As for things like this pull request specifically, I always think it's a good idea to make something as configurable as possible via the command line, it removes the need for users working on an edge case to have to patch the source and compile themselves.

The reason I am packaging this for void by the way, is because the builds of MultiMC do not function on it, as they point to libpng12 while Void currently has libpng18, meaning users would all have to compile it for themselves anyway.

peterix:
I decided to make this into a more radical change - removing 'static data' entirely and 'root' usage for the notifications system.

Issue #1355, opened 2015-12-16
[Linux] [Dev-Version] Couldn't set pixel format


mkg20001:
My Minecraft crashes:
org.lwjgl.LWJGLException: X Error - disp: 0x7f9cf880f1a0 serial: 143 error: GLXBadContext request_code: 155 minor_code: 4

http://paste.ee/p/cwT65

Is there a fix?

wlhlm:
```
libGL error: unable to load driver: r600_dri.so
libGL error: driver pointer missing
libGL error: failed to load driver: r600
libGL error: unable to load driver: swrast_dri.so
libGL error: failed to load driver: swrast
```

It looks like you're missing an accelerated graphics driver.

[Here's](https://bugs.mojang.com/browse/MC-1384) an MC bugreport.

What GPU do you have?

mkg20001:
the problem - it worked in a previously installed MultiMC dev version

mkg20001:
Graphics:  Card: Advanced Micro Devices [AMD/ATI] Broadway PRO [Mobility Radeon HD 5850]
           Display Server: X.Org 1.17.2 drivers: ati,radeon (unloaded: fbdev,vesa) Resolution: 1600x900@59.98hz
           GLX Renderer: Gallium 0.4 on AMD JUNIPER (DRM 2.43.0, LLVM 3.6.2) GLX Version: 3.0 Mesa 11.0.2

peterix:
How do you run MultiMC?

mkg20001:
by desktop shortcut - on ubuntu - with the ubuntu debian package

peterix:
Right. Run 'env' in a terminal and post what that prints out.

Then add 'env' as a pre-launch command and see what that prints.

mkg20001:
```
XDG_VTNR=7
XDG_SESSION_ID=c2
CLUTTER_IM_MODULE=xim
XDG_GREETER_DATA_DIR=/var/lib/lightdm-data/maciej
COMP_WORDBREAKS= 	
"'><;|&(:
GPG_AGENT_INFO=/tmp/gpg-iBsFIj/S.gpg-agent:7952:1
TERM=xterm-256color
SHELL=/bin/bash
VTE_VERSION=4002
WINDOWID=62914570
UPSTART_SESSION=unix:abstract=/com/ubuntu/upstart-session/1000/7615
GNOME_KEYRING_CONTROL=
GTK_MODULES=unity-gtk-module
USER=maciej
[...]

PR #1359, opened 2015-12-17
GH-1355 Fix MMC env propagating to MC after update

SoniEx2:
This closes GH-1355


SoniEx2:
So what was happening is that `${VAR:-${VAR2}}` uses the value of VAR2 if VAR is either unset **or empty**. `${VAR-${VAR2}}` uses VAR2 if VAR is unset.

PR #1419, opened 2016-01-20
Do not ask to overwrite existing file twice.

iarspider:
When exporting an instance to an existing file, the user is asked to confirm overwrite twice:
first by QFileDialog, then by the app itself.

### DIFF
--- a/application/dialogs/ExportInstanceDialog.cpp
+++ b/application/dialogs/ExportInstanceDialog.cpp
@@ -386,7 +386,7 @@ bool ExportInstanceDialog::doExport()
 
 	const QString output = QFileDialog::getSaveFileName(
 		this, tr("Export %1").arg(m_instance->name()),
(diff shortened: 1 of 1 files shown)


02JanDal:
Nice catch, thanks for contributing!

I wonder if it'd be better to leave the confirmation dialog to QFileDialog though, and remove our code instead. If not you should at least replace `NULL` by `nullptr`.

iarspider:
I'll leave this decision to you instead :) But, just in case, replaced NULL with nullptr.

peterix:
One thing. As this has no github issue associated with it, the commite message should be something along the lines of:
```
NOISSUE Do not ask to overwrite existing file twice when exporting instances
```

This is nice and descriptive and won't keep people guessing when they see the changelog during the develop version update.

peterix:
Also, keep it to one commit. Therefore, squash :)

PR #1610, opened 2016-07-06
Make Forge installs only use newer list to fix older version downloads.

FyberOptic:
URLs in the old Forge version list are no longer valid, so you can't install Forge at all on older versions of the game (such as 1.2.5).  So I removed parsing of the old list and modified the newer list parser to discover "client" tags for the older Forge versions so that they'll work properly.

I didn't notice any issues from making this change, though there may have been a reason that this older list was being kept around that I'm not aware of.

Some quick reference links if you want to compare:

Older Forge version list: 
  http://files.minecraftforge.net/minecraftforge/json
Newer Forge version list: 
  http://files.minecraftforge.net/maven/net/minecraftforge/forge/json

URL MultiMC tries to download Forge for 1.2.5 at: 
  http://files.minecraftforge.net/minecraftforge/minecraftforge-client-3.4.9.171.zip
Where as the proper URL now is:
  http://files.minecraftforge.net/maven/net/minecraftforge/forge/1.2.5-3.4.9.171/forge-1.2.5-3.4.9.171-client.zip

### DIFF
--- a/api/logic/minecraft/forge/ForgeVersionList.cpp
+++ b/api/logic/minecraft/forge/ForgeVersionList.cpp
@@ -148,116 +148,6 @@ bool ForgeListLoadTask::abort()
 	return listJob->abort();
 }
 
-bool ForgeListLoadTask::parseForgeList(QList<BaseVersionPtr> &out)
-{
-	QByteArray data;
-	{
-		auto filename = listDownload->getTargetFilepath();
-		QFile listFile(filename);
-		if (!listFile.open(QIODevice::ReadOnly))
-		{
-			return false;
-		}
-		data = listFile.readAll();
-		listDownload.reset();
-	}
-
-	QJsonParseError jsonError;
-	QJsonDocument jsonDoc = QJsonDocument::fromJson(data, &jsonError);
-
-	if (jsonError.error != QJsonParseError::NoError)
-	{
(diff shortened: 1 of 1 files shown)


peterix:
Cool!

Just delete the dead code though. If it's ever needed, it can be pulled from git history.

peterix:
Tested this with 1.5.2 and 1.6.1, because those are the versions where forge went from jar mod to what it is now.

1.5.2 seems to work as it should - using a jar mod.

Some old 1.6.1 forge versions seem to use a jar mod, while they should be using the installer... Possibly they just weren't working right before anyway though.

The 1.6.* versions need to be looked at.

FyberOptic:
I looked in the version json and saw that 1.6.1 Forge versions such as 749 and 571 are listed as "zip" types for the "universal" archive.  It's like this even in the older json as well.  That's likely causing MultiMC to treat it as a jar mod.  You can then notice on 753 that it's then listed as a "jar" type, which I assume is then being handled the modern way.

This seems to just be normal behavior based on the versions of Forge they provide.

peterix:
Alright, squash the commits and change the commit message to something like this:
```
GH-1610 Make Forge installs only use newer list to fix older version downloads
```

peterix:
Hm... I should do something about travis.ci ... it broke.

PR #1671, opened 2016-09-26
Create MultiMC erreur 1

Kavexx:
I think I have drop this issue in a wrong place


peterix:
Yes, please use the Upload button (it's on the log view) and post a new issue with a link to the uploaded log.

PR #1703, opened 2016-11-02
Auto Optfine Intergation #IDEA:

StealthyExpertX:
Optfine installing that would match the forge mod loader downloader and installer system.


StealthyExpertX:
I know this has been talked about before but I'm wondering if its possible now if there has been any new Developments or break troughs that now make this now accomplishable?

peterix:
Just what is going on here? Please use a simple issue, not a pull request from a very old dead branch.

That aside, no - not yet.

PR #1749, opened 2016-12-12
BUILD: Remove extraneous " from CreateServerResourcePacksFolder.cpp

tsmock:


peterix:
This is clearly an error, but it's kinda funny that it builds on all four platforms for me... What are you building with?

tsmock:
It does build, but I was looking through the build log for errors (it doesn't actually fail to build).
I'm going after low hanging fruit first, before going on to more extensive errors -- I've had some issues on Ubuntu 16.10.

peterix:
Alright. The 5.6 Qt build on travis is just broken automation.

The Ubuntu repo with Qt 5.6 has renamed for a minor version bump. Plus, all builds are actually done with 5.4.x - So, It's all OK.

PR #1818, opened 2017-02-18
Update LoginDialog.cpp

32364727:


peterix:
I don't see what this is supposed to accomplish. It just creates unusable broken accounts.

PR #1859, opened 2017-04-19
Add missing <functional> include.

Ristovski:
Compiling under Linux fails because of a missing include directive
Ref:
http://en.cppreference.com/w/cpp/utility/functional/bind
http://en.cppreference.com/w/cpp/utility/functional/placeholders


peterix:
Alright, merged :)

PR #1863, opened 2017-04-22
NOISSUE Update COPYING.MD Formatting

flcmc:


peterix:
Thank you. I haven't really noticed because I originally used a different markdown thingy.

PR #1870, opened 2017-04-26
Fix wrapper command with arguments

AffeAli:
Wrapper commands with arguments were always failing because the command coudn't be found.


AffeAli:
Fixed with 7acf1998eb0683860dc21fc740f1e6bd42f54169.

peterix:
I completely missed this PR. Sorry.

I implemented something similar from scratch yesterday after someone else reported the same problem.

PR #1969, opened 2017-08-27
NOISSUE Update Instance-Version Help Page

flcmc:
The page "Instance-Versions" doesn't exist, it is called "Instance-Version"


peterix:
Can you rebase your commit on the current develop?

PR #1981, opened 2017-09-04
Create I don't know what the problem is. Can some help?

ZCrafter:
Please can someone help!


Forkk:
I think you meant to report an issue, which you can find [here](https://github.com/MultiMC/MultiMC5/issues). However, this looks like it might not be a problem with MultiMC itself, so instead, you should probably [ask for help in our discord](https://discord.gg/ph4K86).

PR #2041, opened 2017-11-17
NOISSUE Implement support for mpticket in classic

MinecraftZuriki:


libraries/launcher/org/multimc/LegacyFrame.java
```diff
@@ -48,12 +51,54 @@ public void start ( Applet mcApplet, String user, String session, int winSizeW,
         try {
             appletWrap = new Launcher( mcApplet, new URL ( "http://www.minecraft.net/game" ) );
         } catch ( MalformedURLException ignored ) {}
+
```
peterix: Add a comment about why this is here.
Something along the lines of:
```
// this allows connecting to classic servers using a system level protocol handler
// that writes the mpticket file and triggers the launch externally
```

MinecraftZuriki: Done.

peterix: Alright, looks good now :)

PR #2103, opened 2018-01-13
Take CMAKE_INSTALL_PREFIX into account in MULTIMC_JARS_LOCATION

freundlich:
When installing MultiMC with the lin-system layout and specifying an
install prefix that is not the empty string, then MultiMC looks for its
Jars in the wrong location. Fix this by appending CMAKE_INSTALL_PREFIX.

### DIFF
--- a/application/CMakeLists.txt
+++ b/application/CMakeLists.txt
@@ -404,7 +404,7 @@ elseif(MultiMC_LAYOUT_REAL STREQUAL "lin-system")
 
 	MESSAGE(STATUS "Compiling for linux system with ${MultiMC_SHARE_DEST_DIR} and MULTIMC_LINUX_DATADIR")
 	set_target_properties(MultiMC PROPERTIES OUTPUT_NAME ${MultiMC_APP_BINARY_NAME})
-	target_compile_definitions(MultiMC PRIVATE "-DMULTIMC_JARS_LOCATION=/${MultiMC_SHARE_DEST_DIR}/jars" "-DMULTIMC_LINUX_DATADIR"
+	target_compile_definitions(MultiMC PRIVATE "-DMULTIMC_JARS_LOCATION=${CMAKE_INSTALL_PREFIX}/${MultiMC_SHARE_DEST_DIR}/jars" "-DMULTIMC_LINUX_DATADIR"
 	)
 
 	# install as bundle with no dependencies included


srakitnican:
`CMAKE_INSTALL_PREFIX` Should be `/usr` by default (It is actually `/usr/local` if not set), that means `MultiMC_SHARE_DEST_DIR` (and the rest `MultiMC_*`) should be changed to not include usr/ prefix.

I am not exactly sure, but I think new generation package managers could benefit from this as well.

srakitnican:
Something like the following, also adds the library suffix if `LIB_SUFFIX` is set as found on some distributions like Fedora.

```diff
--- /tmp/CMakeLists.txt	2018-01-14 12:08:30.395594381 +0100
+++ /tmp/CMakeLists.txt.patched	2018-01-14 12:07:16.604106019 +0100
@@ -393,9 +393,13 @@
 	install(PROGRAMS package/linux/MultiMC DESTINATION ${BUNDLE_DEST_DIR})
 
 elseif(MultiMC_LAYOUT_REAL STREQUAL "lin-system")
-	set(MultiMC_BINARY_DEST_DIR "usr/bin" CACHE STRING "Relative path from packaging root to the binary directory")
-	set(MultiMC_LIBRARY_DEST_DIR "usr/lib" CACHE STRING "Relative path from packaging root to the library directory")
-	set(MultiMC_SHARE_DEST_DIR "usr/share/multimc" CACHE STRING "Relative path from packaging root to the shared data directory")
+	if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+          SET(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "Base directory for executables and libraries" FORCE)
+       endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+
+	set(MultiMC_BINARY_DEST_DIR "${CMAKE_INSTALL_PREFIX}/bin" CACHE STRING "Relative path from packaging root to the binary directory")
+	set(MultiMC_LIBRARY_DEST_DIR "${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}" CACHE STRING "Relative path from packaging root to the library directory")
+	set(MultiMC_SHARE_DEST_DIR "${CMAKE_INSTALL_PREFIX}/share/multimc" CACHE STRING "Relative path from packaging root to the shared data directory")
 	set(MultiMC_APP_BINARY_NAME "multimc" CACHE STRING "Name of the MultiMC binary for the purposes of linux packaging")
 	set(JARS_DEST_DIR "${MultiMC_SHARE_DEST_DIR}")

```

peterix:
OK. I clicked a github button again and it broke things by adding merge commits -- again...

peterix:
Cherry-picked the change.

@srakitnican  I'm going to make some larger changes to the build system and include that. Looks good.

peterix:
Ok, this was extremely messy... Basically everything broke.

A few force-pushes and some fixing of the AUR package later, this is how this looks:
https://github.com/MultiMC/MultiMC5/blob/e8336babad17741cfaacec7c401db28cdabed623/application/CMakeLists.txt#L395-L416

srakitnican:
What exactly did not work? I tested my changes and it was fine for me. With my changes applied, change from @freundlich is redundant.

I believe this would result in jars being installed in `/usr/usr/share/multimc`:
https://github.com/MultiMC/MultiMC5/blob/d0e58acd843fcc04463497e9ffbbcab0dee72d55/application/CMakeLists.txt#L402,L411

srakitnican:
> I agree that removing usr from all of these paths is the right thing to do. However, I think that changing the default value of CMAKE_INSTALL_PREFIX isn't really needed and might add additional confusion.

Well I did it like that to not make any further noticeable changes in terms where it installs. By default prefix is `/usr/local`, without this it would be required to manually specify `-DCMAKE_INSTALL_PREFIX=/usr` if you want to install to `/usr`. I am fine with `/usr/local` being default, but I didn't want to suggest to break that for existing users.

srakitnican:
> I believe this would result in jars being installed in `/usr/usr/share/multimc`:
https://github.com/MultiMC/MultiMC5/blob/d0e58acd843fcc04463497e9ffbbcab0dee72d55/application/CMakeLists.txt#L402,L411

I did some tests and apparently I was wrong. It seems CMake automatically combines multiple variables into one, not sure how this works.

What it also does is that it prepends `CMAKE_INSTALL_PREFIX` to all files with relative path in install command so specifying it manually is not necessary.


> DESTINATION
> Specify the directory on disk to which a file will be installed. If a full path (with a leading slash or drive letter) is given it is used directly. If a relative path is given it is interpreted relative to the value of the `CMAKE_INSTALL_PREFIX` variable. The prefix can be relocated at install time using the `DESTDIR` mechanism explained in the `CMAKE_INSTALL_PREFIX` variable documentation.

I have found that the following is all that is needed to MultiMC 0.6.0.
```diff
--- application/CMakeLists.txt	2017-12-31 19:39:46.000000000 +0100
+++ application/CMakeLists.txt.patched	2018-01-17 09:21:28.018546384 +0100
@@ -393,9 +393,12 @@
 	install(PROGRAMS package/linux/MultiMC DESTINATION ${BUNDLE_DEST_DIR})
 
 elseif(MultiMC_LAYOUT_REAL STREQUAL "lin-system")
-	set(MultiMC_BINARY_DEST_DIR "usr/bin" CACHE STRING "Relative path from packaging root to the binary directory")
-	set(MultiMC_LIBRARY_DEST_DIR "usr/lib" CACHE STRING "Relative path from packaging root to the library directory")
-	set(MultiMC_SHARE_DEST_DIR "usr/share/multimc" CACHE STRING "Relative path from packaging root to the shared data directory")
+	if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+		set(CMAKE_INSTALL_PREFIX "/usr" CACHE PATH "Base directory for executables and libraries" FORCE)
+	endif()
+	set(MultiMC_BINARY_DEST_DIR "bin" CACHE STRING "Relative path from packaging root to the binary directory")
+	set(MultiMC_LIBRARY_DEST_DIR "lib${LIB_SUFFIX}" CACHE STRING "Relative path from packaging root to the library directory")
+	set(MultiMC_SHARE_DEST_DIR "share/multimc" CACHE STRING "Relative path from packaging root to the shared data directory")
 	set(MultiMC_APP_BINARY_NAME "multimc" CACHE STRING "Name of the MultiMC binary for the purposes of linux packaging")
 	set(JARS_DEST_DIR "${MultiMC_SHARE_DEST_DIR}")
 ```

Also even setting the `CMAKE_INSTALL_PREFIX` is not needed since it was probably broken before as well since the relative path was used, thus if  `CMAKE_INSTALL_PREFIX` was not set it would install it to `/usr/local/usr...`. So with this new change we are breaking this use case anyway so we might as well not set the `CMAKE_INSTALL_PREFIX` by default as suggested by @freundlich . Users will have to adjust their `CMAKE_INSTALL_PREFIX` anyway so we might as well set the sane defaults since `/usr` is meant to be used by packages. `/usr/local` is the right place for manually compiled stuff.

```diff
--- application/CMakeLists.txt	2017-12-31 19:39:46.000000000 +0100
+++ application/CMakeLists.txt.patched	2018-01-17 09:21:28.018546384 +0100
@@ -393,9 +393,12 @@
 	install(PROGRAMS package/linux/MultiMC DESTINATION ${BUNDLE_DEST_DIR})
 
 elseif(MultiMC_LAYOUT_REAL STREQUAL "lin-system")
-	set(MultiMC_BINARY_DEST_DIR "usr/bin" CACHE STRING "Relative path from packaging root to the binary directory")
-	set(MultiMC_LIBRARY_DEST_DIR "usr/lib" CACHE STRING "Relative path from packaging root to the library directory")
-	set(MultiMC_SHARE_DEST_DIR "usr/share/multimc" CACHE STRING "Relative path from packaging root to the shared data directory")
+	set(MultiMC_BINARY_DEST_DIR "bin" CACHE STRING "Relative path from packaging root to the binary directory")
+	set(MultiMC_LIBRARY_DEST_DIR "lib${LIB_SUFFIX}" CACHE STRING "Relative path from packaging root to the library directory")
+	set(MultiMC_SHARE_DEST_DIR "share/multimc" CACHE STRING "Relative path from packaging root to the shared data directory")
 	set(MultiMC_APP_BINARY_NAME "multimc" CACHE STRING "Name of the MultiMC binary for the purposes of linux packaging")
 	set(JARS_DEST_DIR "${MultiMC_SHARE_DEST_DIR}")
 ```

PR #2113, opened 2018-01-18
GH-2103 Make /usr/local the default prefix for lin-system

srakitnican:
/usr/local is a sane default since /usr is meant to be used by packages.

Better to fix it now since we are already redesigning then later. Users would have to adjust their variables anyway.


srakitnican:
Travis script would need an update as well.

peterix:
No, travis is not making packages. It builds the code and runs unit tests. The actual builds are done on the [buildbot builders](http://ci.multimc.org/waterfall), and these build the portable versions, push them to the website and the s3 update repo.

PR #2171, opened 2018-02-26
FTB Support

Janrupf:
Started to integrate FTB into MultiMC


MultiMC-Repo.kdev4
```diff
@@ -0,0 +1,4 @@
+[Project]
```
peterix: Please do not include this file

PR #2231, opened 2018-04-07
Updated ftb pack chooser to auto import icons and fixed to big version list

Janrupf:


peterix:
Cherry-picked and fixed (to make it build on ci).

PR #2327, opened 2018-07-01
Fix MultiMC spelling in .desktop file

Ameliorate:
At some point, the multimc binary was renamed from `multimc` to `MultiMC`, causing the .desktop file to break and not show up in the system applications menu.


NikkyAI:
at least on the official package in the aur https://aur.archlinux.org/packages/multimc-git/
the binary is still all lowercased, but the launcher is named `MultiMC`

Ameliorate:
Oh, I see what is causing the issue. Nix packages multimc using a build script that outputs `MultiMC`, however the AUR uses the cmake setup and outputs `multimc`. It seems this is completely a downstream issue.

PR #2349, opened 2018-07-23
Feature/ftb pack codes

Janrupf:


peterix:
I manually re-created a single commit that does this, with some small changes (making the private pack manager a normal object, held by the FTB page).

In any case, merged.

PR #2370, opened 2018-08-11
add "minecraft" keyword to .desktop entries

StripedMonkey:
This is pretty much just a QoL change that lets you search "minecraft" and have multiMC also pop up in search menus. I also added the `Keywords` key to the .desktop entry in the Linux portion, its considered a standard key so it shouldn't cause any issues from DE to DE.


peterix:
Merged manually using cherry-pick to avoid merge commits.

PR #2447, opened 2018-11-02
NOISSUE use https more widely

ChrisLane:
I was alerted to the fact that HTTPS might not be the default for URLs in MultiMC after seeing 9b74e73ad300404049f476d2a79e5b63f4665fee.

HTTPS should be used wherever possible to reduce security concerns and avoid HTTP 3xx redirection.
This PR addresses these concerns.


api/logic/net/URLConstants.h
```diff
@@ -25,15 +25,15 @@ const QString LIBRARY_BASE("libraries.minecraft.net/");
...
-const QString FORGE_LEGACY_URL("http://files.minecraftforge.net/minecraftforge/json");
-const QString FORGE_GRADLE_URL("http://files.minecraftforge.net/maven/net/minecraftforge/forge/json");
-const QString MOJANG_STATUS_URL("http://status.mojang.com/check");
-const QString MOJANG_STATUS_NEWS_URL("http://status.mojang.com/news");
-const QString LITELOADER_URL("http://dl.liteloader.com/versions/versions.json");
+const QString FORGE_LEGACY_URL("https://files.minecraftforge.net/minecraftforge/json");
+const QString FORGE_GRADLE_URL("https://files.minecraftforge.net/maven/net/minecraftforge/forge/json");
+const QString MOJANG_STATUS_URL("https://status.mojang.com/check");
+const QString MOJANG_STATUS_NEWS_URL("https://status.mojang.com/news");
+const QString LITELOADER_URL("https://dl.liteloader.com/versions/versions.json");
```
Janrupf: This URL seems to not work with https.

peterix: And sometimes the URLs are obsolete and no longer used... Maybe they should be removed instead, if there's no remaining use.

ChrisLane: > This URL seems to not work with https.

Ah yes, checked it served a valid cert but didn't see the 404.

ChrisLane: > And sometimes the URLs are obsolete and no longer used... Maybe they should be removed instead, if there's no remaining use.

Unused URLs from what I can see:
RESOURCE_BASE
FORGE_LEGACY_URL
FORGE_GRADLE_URL
MOJANG_STATUS_NEWS_URL
LITELOADER_URL

PR #2459, opened 2018-11-09
Fix a few usability issues

### LINKED ISSUE #1552: Suggestion: Auto-gray Permgen option for Java 8+ or allow individual activation of each memory option
new versions of java don't HAVE a permgen option anymore, and the VM puts a complaint when you supply it. No harm done, its more of an extremely minor irritation than anything super important..

Also wouldn't hurt to have an advanced panel with some of the other memory options, and some more common garbage collector and thread model options in a checklist.

Obviously that is achievable already with the arguments box, but it'd be neat.

### LINKED ISSUE #2144: The version selection dialog should highlight the current installed version, if there is one
* -> Maybe an arrow?
* Or put '#installed' somewhere? 
* Or maybe just select it by default?

### LINKED ISSUE #2384: Dragging a mod with the same filename into the Mods screen doesn't replace the existing copy
System Information
-----------------------------
MultiMC version: 0.6.2-develop-1245 on win32

Operating System: Windows 7

Summary of the issue or suggestion:
----------------------------------------------
When clicking and dragging a mod file into the Loader Mods list, when a mod already exists with the same filename, nothing happens.  

What should happen:
------------------------------
I'd love it if the newly dragged-in copy would replace the old one!

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. Procure some mod jar file, and click-drag it into the Loader Mods list.

2. Change the contents of it, but give it the same name.

3. Drag the new file into the Loader Mods list.

4. Nothing happens (the old file is still in your instance folder)

Suspected cause:
---------------------------
Fairy gremlins

Logs/Screenshots:
----------------------------
Nothing out of the ordinary.

Additional Info:
---------------------------
This would be great for mod development - I manage my (outside-dev environment) test instances with MultiMC, and it would be nice to update the copy of a mod in my instance without having to `View Folder` and fiddle with files manually. The version number doesn't always change, meaning the mod's filename isn't always different.

Just a little thing.

### LINKED ISSUE #2452: [Feature Request] Confirmation box for the Delete Group button
Upon right clicking inside of any group your confronted with two buttons very close to one another, Create instance and Delete group. Because of the fact that there is no confirm dialog for this delete group button, I've ended up deleting one of my groups while instead just trying to create a new instance.

I recommend either some kind of confirm dialog for the delete group button or to make the option only appear upon clicking a specific interface element, such as the group title or something.

Janrupf:
I fixed some small usability issues.
This PR:

- closes #2452 
- fixes #2384 
- closes #2144 
- closes #1552 

(Note: Some ideas were adjusted a bit, for example perm gen will not be displayed and not grayed out since it doesn't make sense to just gray it out)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@ CMakeLists.txt.user
 CMakeLists.txt.user.*
 /.project
 /.settings
+/.idea
+cmake-build-*/
 
 # Build dirs
 build
--- a/api/logic/minecraft/SimpleModList.cpp
+++ b/api/logic/minecraft/SimpleModList.cpp
@@ -146,6 +146,8 @@ bool SimpleModList::installMod(const QString &filename)
     if (type == Mod::MOD_SINGLEFILE || type == Mod::MOD_ZIPFILE || type == Mod::MOD_LITEMOD)
     {
         QString newpath = FS::PathCombine(m_dir.path(), fileinfo.fileName());
+        if(QFile::exists(newpath))
+            QFile::remove(newpath);
         if (!QFile::copy(fileinfo.filePath(), newpath))
             return false;
         FS::updateTimestamp(newpath);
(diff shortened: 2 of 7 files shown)


peterix:
Merged with cherry picks and additional tweaks.

PR #2572, opened 2019-03-19
Fixes #2515 selectedInstance was wrong if instance was deleted manually.

### LINKED ISSUE #2515: Deleting an instance externally with deletion confirmation open deletes another pack
System Information
-----------------------------
MultiMC version: 0.6.4-1291

Operating System: Windows 10 (10.0.16299)

Summary of the issue or suggestion:
----------------------------------------------
Upon deleting a pack, if the deletion confirmation is accepted after the pack is deleted externally, another pack will be deleted in its place.

What should happen:
------------------------------
An error or similar event preventing the deletion of the "innocent" pack.

Steps to reproduce the issue:
-------------------------------------------------------------
1. Open the deletion confirmation for a pack

2. Delete the pack externally, then confirm the dialog

3. Another pack will be deleted in its place.

Suspected cause:
---------------------------
Something in InstanceList::deleteInstance, but I can't seem to find the bug. Unless getInstanceById returning an empty InstancePtr can somehow coincide with an existing instance. 

Additional Info:
---------------------------
Looking through my logs, I don't see anything that has the debug strings found in InstanceList::deleteInstance, unless it has been overwritten; I've opened MultiMC a few times since the scare (I had a backup of the instance thankfully)

GeorgeWilson96:


peterix:
Why did you delete this?

Issue #2551, opened 2019-02-28
MultiMC does not remove newlines from servers menu


Zekken01:
![image](https://user-images.githubusercontent.com/19770864/53594228-68a82b80-3b68-11e9-85f1-4413a19dbef4.png)

![image](https://user-images.githubusercontent.com/19770864/53594149-3991ba00-3b68-11e9-9e78-defd3e2945a5.png)

This can be an issue when copying server IPs from websites.

PR #2573, opened 2019-03-19
GH-2551 Trim server name and IP strings

robotbrain:
Trims the name and ip fields from the add server dialog so that trailing whitespace (especially newlines) don't render oddly in the minecraft client.


peterix:
This trims the strings parsed from NBT. I think it should trim the strings saved to NBT instead.

PR #2651, opened 2019-05-11
Adjust instance importing so that it can use a hybrid approach.

Tsudico:
Start with a MultiMC instance and add Flame mods to it. This will allow using
Fabric mod loader MultiMC instance generation (https://fabricmc.net/use/) with Flame manifest.json and overrides.

If both types are found, it will use the root of MultiMC for both. If roots should be found separately I can add that capability.

### DIFF
--- a/api/logic/InstanceImportTask.cpp
+++ b/api/logic/InstanceImportTask.cpp
@@ -83,7 +83,14 @@ void InstanceImportTask::processZipPack()
     QString mmcFound = MMCZip::findFolderOfFileInZip(m_packZip.get(), "instance.cfg");
     QString flameFound = MMCZip::findFolderOfFileInZip(m_packZip.get(), "manifest.json");
     QString root;
-    if(!mmcFound.isNull())
+    if(!mmcFound.isNull() && !flameFound.isNull())
+    {
+        // process as MultiMC instance with Flame mods
+        qDebug() << "MultiMC:" << mmcFound << ", Flame:" << flameFound;
+        root = mmcFound;
+        m_modpackType = ModpackType::Hybrid;
+    }
+    else if(!mmcFound.isNull())
     {
         // process as MultiMC instance/pack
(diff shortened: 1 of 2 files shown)


Shnupbups:
I would absolutely love this.

I have a Fabric modpack that would, ideally, be distributed as a MultiMC instance, but I don't want to redistribute people's mods, so I'd like it to download the mods from CurseForge. Right now, with Flame, it *can* do that, but in doing so I can no longer have the instance come with Fabric Loader, meaning the whole thing is for nought.

With a hybrid approach, everything would be much nicer.

nift4:
What is Flame? Google but not found anything.

Tsudico:
I believe it's how MultiMC refers to Curseforge's modpack format. I know it works with loading mods from Curseforge.

peterix:
The mmc pack format should never download from curse. These should always be two distincrt formats with zero overlap. Allowing overlap would be a terrible thing to do.

1. If you allow that, you allow to have curseforge to have files that have no chance to work with their client even though they look like they should, and it would attempt to import them and fail.
As much as I dislike it (curse), that sounds horrible to me.
None of the potential outcomes of it are desirable. It could be seen as an attack. It could lead to someone from curse actually trying to implement MultiMC specific things. I do not want any of that.
2. It further expands on what curse is allowed into, in MultiMC.
That set should be empty. I allow it only because curse itself is incapable of supporting certain platforms
It is a business. One that I am not part of, and one that does not need my support or cooperation.

phit:
just fyi
1. that is already the case, they are a hosting platform first and host a lot of things their launcher can't use

peterix:
If it is just storage, then let's treat it just as storage and stop using the metadata formats it comes with.

peterix:
In other words.
Fabric: OK, but only done properly. Not this. Ever.
Anything curse: Over my dead body, and I've been done discussing that years ago.

Issue #2499, opened 2018-12-28
Proxies should apply instantly


kb-1000:
System Information
-----------------------------
MultiMC version: latest develop (0.6.2-1283)

Operating System: Windows 10

Summary of the issue or suggestion:
----------------------------------------------
If I change proxy settings, I need to restart MultiMC. This is pretty confusing, because it's noted nowhere and people could assume proxies don't work (like in #1967).

What should happen:
------------------------------
The change instantly applies

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. Setup a non-existing proxy and restart MultiMC

2. Remove the proxy

3. Try to download anything

Suspected cause:
---------------------------

Logs/Screenshots:
----------------------------
[//]: # (Please refer to https://github.com/MultiMC/MultiMC5/wiki/Log-Upload for instructions on how to attach your logs.)

Additional Info:
---------------------------

PR #2703, opened 2019-06-23
GH-2499 Apply proxy settings immediately

Janrupf:
Proxy settings are applied immediately.


peterix:
This is a bit problematic, because it will not be applied to existing requests. I'm not sure what to do about that though.

Basically, if you have a broken proxy that leads to long timeouts, and turn it off, a lot of things in MultiMC will keep being unresponsive for a while.

Maybe fixing that is something to do in the future :)

PR #2705, opened 2019-06-23
GH-2515 Save instance ID before display dialog

Janrupf:
Deleting an instance externally no longer deletes the next instance.


peterix:
Good stuff!

PR #2707, opened 2019-06-23
Feature/rename folders with instances

Janrupf:
When renaming instances, there is now a checkbox to also rename the folder.


peterix:
Sorry :(

Issue #689, opened 2014-12-22
MultiMC creates imgur albums even when uploading single screenshots


peterix:

peterix:
Moved to workflowy

PR #2708, opened 2019-06-23
GH-689 Don't create album for single screenshot

Janrupf:
MultiMC no longer creates albums for single screenshots.


peterix:
Tested and works :+1:

PR #2709, opened 2019-06-23
Feature/limit instance folder name length

Janrupf:
This PR limits the instance folder name length to 240 chars. This should be a sane limit working on all platforms.


peterix:
I don't believe this is correct. It should generate a truncated name in the first place, not mess with the original InstID and copypaste a subset of it.

That is, change `DirNameFromString` or make a specialized copy of it.

peterix:
This is a bit crazy... like, absolute paths longer than 240 characters are a normal thing that shouldn't ever fail?

Janrupf:
Yeah, I hope this never becomes an issue again. Over the years things have changed a bit, in case someone ever runs into this, here is some special information for windows: https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#maximum-path-length-limitation

PR #2907, opened 2019-11-03
Add support for creating RPM Packages

sambhavsaggi:
Something simple to create RPM packages for MultiMC. Can be build and only has to be installed once.

TODO: Fix the dependencies

Note that most of this was taken from the Ubuntu repository.


peterix:
Would it be possible to not use the source tarball? Binary blobs like that do not belong in git.

application/package/fedora/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+# How to build this?
+You need to install the following:
+
+```
+sudo dnf install fedora-packager fedpkg fedrepo_req copr-cli
+```
+
+Copy the `rpmbuild` folder to `rpmbuild_0.6.7` and then run:
+```
```
spannerman79: Should use scripting to auto grab the tarball
See https://github.com/MultiMC/MultiMC5/blob/53ad5cb436ea4bffb93c7d26a1be43690a21bf0d/application/package/ubuntu/multimc/opt/multimc/run.sh#L5..L19 for a reference

sambhavsaggi:
The tarball is *not* the MultiMC code. It simply contains the package for the RPM file (desktop files, auto-install script for actual tarball).

Sorry for the late reply. I was busy with schoolwork.

ProjectSynchro:
This one came first,
It didn't appear that any progress was happening so I created another PR.

ElXreno:
How about use it? https://copr.fedorainfracloud.org/coprs/elxreno/multimc

peterix:
Keep this open. I want to actually look at it at some point, not forget it existed.

sambhavsaggi:
Ah. I was under the impression that rpm support had already been added https://github.com/MultiMC/MultiMC5/blob/develop/application/package/rpm/MultiMC5.spec

peterix:
Well, something has been added. I'd like to actually compare and use the best pieces of both.

peterix:
There is no RPM on the website yet.

sambhavsaggi:
Why was this closed?

jamierocks:
Likely because an rpm package was merged at some point.

https://github.com/MultiMC/MultiMC5/issues/3955 does exist to establish whether anything from here should be merged though :)

PR #2990, opened 2020-02-04
feat: add --import command-line option

### LINKED ISSUE #2988: Flag to import zip
Suggestion for a flag that automatically imports, or opens the import zip screen and enters the provided path (for the user to confirm).

I wrote a program to read twitch's CCIP files and download the zip, but there's no easy way to tell multimc to import that zip automatically.

OverMighty:
This pull request adds an `--import` command-line option (and an `-I` short option) which, when specified, makes the "Import from zip" dialog automatically open when the main window is shown, with the URL field's value set to the URL specified as command-line argument.

In order to add this feature, I had to **make `MainWindow::droppedURLs(QList<QUrl>)` public.** I don't know if you're okay with that or not.

Closes #2988

### DIFF
--- a/application/MainWindow.h
+++ b/application/MainWindow.h
@@ -57,6 +57,8 @@ public:
     void checkInstancePathForProblems();
 
     void updatesAllowedChanged(bool allowed);
+
+    void droppedURLs(QList<QUrl> urls);
 signals:
     void isClosing();
 
@@ -179,8 +181,6 @@ private slots:
      * Runs the DownloadTask and installs updates.
      */
     void downloadUpdates(GoUpdate::Status status);
-
-    void droppedURLs(QList<QUrl> urls);
 
     void konamiTriggered();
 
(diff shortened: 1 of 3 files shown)


peterix:
To do that, you may have to change/customize the protocol in between.

These need to be updated:
https://github.com/MultiMC/MultiMC5/blob/develop/application/MultiMC.cpp#L274-L292
https://github.com/MultiMC/MultiMC5/blob/develop/application/MultiMC.cpp#L844-L863

Ideally, also fix the launch command to not be just a name of something. Make it `launch ID` instead of `ID` like it is now.

OverMighty:
Ok, I will work on this.

OverMighty:
Done. `activate` messages still tell the main process to show the main window, I replaced the `<id>` messages with `launch <id>`, and I added a new `import <url>` kind of message.

peterix:
Just one thing: I'll squash/rebase this to make it follow the commit message format MultiMC uses.

peterix:
Merged.

PR #2996, opened 2020-02-08
Add home dir configuration via environment variable

ChrisLane:
This adds the ability for the MultiMC directory to be configured via an environment variable.
Config preference:
1. Launch argument
2. Environment variable
3. XDG home
4. Fallback dir

### DIFF
--- a/application/MultiMC.cpp
+++ b/application/MultiMC.cpp
@@ -218,11 +218,20 @@ MultiMC::MultiMC(int &argc, char **argv) : QApplication(argc, argv)
     else
     {
 #ifdef MULTIMC_LINUX_DATADIR
-        QString xdgDataHome = QFile::decodeName(qgetenv("XDG_DATA_HOME"));
-        if (xdgDataHome.isEmpty())
-            xdgDataHome = QDir::homePath() + QLatin1String("/.local/share");
-        dataPath = xdgDataHome + "/multimc";
-        adjustedBy += "XDG standard " + dataPath;
+        QString envMultimcHome = QFile::decodeName(qgetenv("MULTIMC_HOME"));
+        if (!envMultimcHome.isEmpty())
+        {
+            dataPath = envMultimcHome;
+            adjustedBy = "Environment variable " + dataPath;
+        }
+        else
+        {
+            QString xdgDataHome = QFile::decodeName(qgetenv("XDG_DATA_HOME"));
+            if (xdgDataHome.isEmpty())
+                xdgDataHome = QDir::homePath() + QLatin1String("/.local/share");
+            dataPath = xdgDataHome + "/multimc";
+            adjustedBy += "XDG standard " + dataPath;
+        }
 #else
         dataPath = applicationDirPath();
         adjustedBy += "Fallback to binary path " + dataPath;


ChrisLane:
I understand that you can configure  this via XDG_DATA_HOME, however, this does not give the same level of flexibility or convenience that a new environment variable would give.

- There would be no way to put the MultiMC root in a directory other than "multimc"
- If you reconfigure the XDG environment variable just for multimc, you've tainted it when compared to your global setting. The variable can no longer be used in future within multimc for its main use as a global setting.
- Configuring XDG_DATA_HOME on every run of multimc would be a hassle, the user would have to set up an alias and even then, that doesn't cover situations where the user might be using a launcher such as rofi. All of these additional situations would have to be configured to have a custom XDG setting.
- The `--dir` argument overrides the XDG_DATA_DIR configuration, it seems suitable that an environment variable should exist to match this behaviour. XDG_DATA_DIR as a standardised location configuration should remain **common** to all programs (as it is supposed to be) instead of hijacking it per-application.

ShayBox:
I disagree that you should be setting global environment variables for single applications, even if it has no side-effects, I would still modify the desktop file to add flags/env like you're supposed to for desktop environments, or for cli environments you would as you said, make an alias to override the command (`alias multimc="multimc --dir $HOME"`), but to each their own I suppose.

wlhlm:
> I disagree that you should be setting global environment variables for single applications, even if it has no side-effects, for example: on a multi-user system, that would break other user's multimc, ...

I think you misunderstood. 'global' was referring to `XDG_*` variables, as in 'global for all applications that understand `XDG_*` variables'. Also, you can use environment variables in as wide, or as narrow of a context as you want. You don't have to specify `MULTIMC_HOME` for all users, you can configure it per user (`.pam_environment`, `.bashrc`, etc.), or even per single invocation of MultiMC (`$ MULTIMC_HOME=/mnt/mc multimc` on the shell, but of course in that case you can also use `XDG_DATA_DIR`).

ShayBox:
Good point, I forgot you can set per user "global" environment variables, either way I guess there is a use for this

peterix:
I'm not sure what this is trying to solve, but maybe start by explaining that.

I'm closing all outstanding PRs in order to make some sweeping changes.

ChrisLane:
It's fairly standard to be able to reconfigure data or config directories for software in linux using environment variables.

MultiMC already supports XDG directories and setting the data directory via an argument, that's great!

The main problem I have is that I want my MultiMC data in my cloud storage directory and to launch MultiMC via a launcher, not from a terminal.
Here are my current options:
1. Redefine `XDG_DATA_HOME` - This would either affect all applications or need redoing each time MultiMC updates if I edit the .desktop file.
2. Use the --dir argument - This would need to be done in the .desktop file and need redoing each time MultiMC updates, alternatively it could be done in an alias and started in the terminal.
3. Create a symlink in the `XDG_DATA_DIR` to the correct location - This is what I currently do. Problem is that it's not a config that can easily be moved between systems via normal dot files

Benefits of having a new environment variable specifically for MultiMC's data:
- Can be used like `MULTIMC_HOME=~/cloud/multimc-dir multimc` without tainting `XDG_DATA_HOME` for wrapper scripts etc. used within MultiMC.
- Unnecessary symlinks don't need to be created to work around the missing environment variable.
- The new environment variable can be placed in a file that's sourced in login scripts to change the data location for MultiMC wherever it's called without the need for updating .desktop files or creating aliases.

Hopefully that's made things a bit clearer. Let me know if anything needs clearing up.

PR #2997, opened 2020-02-08
Add man page doc

ChrisLane:
Adds an asciidoc that can be used to generate a man page. This is additional work so that the environment variable in #2996 can be documented in a man page.

 The AUR PKGFILE will need to be updated to:
 - add `asciidoc` to `makedepends`
 - run `a2x --format manpage`
- install the generated `multimc.1` file at `/usr/share/man/man1/`

### DIFF
--- a/doc/multimc.1.txt
+++ b/doc/multimc.1.txt
@@ -1,0 +1,62 @@
+MULTIMC(1)
+==========
+:doctype: manpage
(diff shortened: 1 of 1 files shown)


peterix:
Eh, why not. That's nice. Just needs integration into the packages.

PR #3071, opened 2020-04-03
Added primitive search feature for mods

louis123456:
I implemented a search and download feature so you can download mods from CurseForge.


louis123456:
I forgot to add a version select so you can select the version of the mod you like... So this is not ready yet

louis123456:
This should be working now.

Janrupf:
You should edit the commit message so they match the commit message scheme in MultiMC

peterix:
I do not want this sort of integration with Twitch.

Issue #3010, opened 2020-02-22
multimc wont start anymore


SevenChords:
System Information
-----------------------------
MultiMC version: 
latest stable

Operating System: 
macos catalina (latest)

Summary of the issue or suggestion: 
the application multimc doesn't start anymore, it bounces in the dock as if it is starting, after that it just sits still in the dock without the dot below it indicating a running application (i normally don't have it in my dock), but it shows up in activity monitor as running, nothing happens anymore, can't click it, can't do anything but restarting my mac or force quitting multimc
----------------------------------------------

What should happen: 
multimc should start normally and open the main window
------------------------------

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. open multimc on macos catalina (latest)

2. try to import new instance from a local zip file

3. repeat step 2 until it freezes with rainbow spinny circle

4. force quit multimc after freeze in step 3

5. try to start multimc again

Suspected cause: 
issues/conflicts with the way macos catalina and multimc handle file/disk access and permissions for said access
---------------------------

peterix:
Apple being Apple and soing stupid things with 'security'...

phit:
this was presumably fixed with https://github.com/MultiMC/MultiMC5/pull/3081
if you are still having issues on the development version please let us know

Issue #2784, opened 2019-08-04
MAC - Freeze


Pseudohuman93:
Hello!

I got an issue with the latest version of MultiMC for Mac - (Mac version - MacOS Mojave - Latest version).

For some reason, every time I try to open the app it freezes for 40 seconds. The same happens when I launch a version of Minecraft. 40 seconds too. After that, everything else works perfectly.

peterix:
AFAIK, this is macOS doing pointless work for security theater reasons. I can't really do anything about it aside from removing features (MultiMC being a portable application), or dropping macOS support altogether.

It is a problem, not sure what I'll do about it yet.

Boundarybreaker:
I'm not entirely sure this is security theater stuff. It freezes the whole screen for a good few seconds every time a new window is opened, even the game window or the new instance window.

Sakata-MC:
does the same crap in windows 10

Heufneutje:
No, this problem is specific to macOS. Create a new issue.

adamyellen:
A workaround for this issue: launch the binary from the terminal.

Boundarybreaker:
…huh. For some reason, this issue just *stopped happening* when I updated to Catalina. I have *no* idea why.

peterix:
It is a macOS bug.

KNOBDev:
Apparently, this only happens when you give MultiMC Accessibility features in Security & Privacy settings.

refactorized:
This is a definite issue, and not one i have seen on any other mac application, portable or not, native or not.  I love MultiMC but it freezes the entire visual surface of the OS everytime I launch it, or go to open files.  It's terrible.

Because i don't see this anywhere else, I have to assume it is at least very fixable (if not trivial) either as an mmc bug or an osx bug with a viable workaround.

refactorized:
I can confirm that disabling Accessibility features fixes this! I don't know why MMC needs those features, and therefore what disabling them might break, but things look good so far.

Ianshogren:
> Apparently, this only happens when you give MultiMC Accessibility features in Security & Privacy settings.

This fixed it, thank you! Do you know why?

Zetabite:
Yes. MacOS security feature is as cancer and bad as Winodws. It is supposed to stop viruses and stuff, but mostly just blocks programs whose developers wont pay those stupid licsenses to get on the good boy list

Boundarybreaker:
[...]

Issue #2598, opened 2019-04-12
Multimc won't open on my mac


31trainman:
System Information Mac os x mojave 10.14.4 
MacBook Pro ( Retina, Mid 2012)
2.6 GHz Intel Core i7
8GB ram
-----------------------------
MultiMC version: 0.6.2-1163

Operating System: Mac os mojave

Summary of the issue or suggestion:
----------------------------------------------
Trying to launch multimc freezes my mac up and only a power off fixes it

What should happen:
------------------------------
Multimc should open normally

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. Try to launch multimc normally

2.

3.

Suspected cause:
---------------------------
Not sure

Logs/Screenshots:
----------------------------
[//]: # https://pastebin.com/ngNYun7X

Additional Info:
---------------------------

Kename:
Same situation in Mojave with a different set-up:
2015 MB Pro 13"
3.1 GHz i7
16 GB
OS 10.14.4

peterix:
Mojave blocks the launch because it checks all the files in the application every time.

I'd call it an OS bug, but Apple will never fix this. It's intentional.

Issue #2579, opened 2019-03-26
MultiMC Crashes when attempting to open Folder from an Instance


madalienist:
System Information
-----------------------------
MultiMC version:0.6.5-1243

Operating System: MacOs 10.14.3

Summary of the issue or suggestion:
----------------------------------------------
I created an instance of a ModPack (FTB Unlimited Reloaded), tried to open the instance folder (or any linked folder) and MultiMC crashes (spinning ball of death).  I wanted to find the packs scripts folder so I could modify a script for the pack.  Tried dozens of times.  Restarted the machine, restarted MultiMC.  Loaded the modpack to make certain all folders were created.  Nothing worked.  

I finally copied my world over (from twitch), copied over a couple of Mods I had added.  Then launched the ModPack.  In the middle it said I was missing a mod and it gave me the option of opening the mod folder.  Wala, i could see the folders.  Added the changed script and the missing mod.  Forgot to add the journeyman folder.  

I want to be able to access the instance folder and add the journey map folder from my Twitch play through.  

What should happen:
------------------------------
Folders should open

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1.Open MUltiMC. 

2.Pick a Folder associated with an Instance

3.Watch it crash.

Suspected cause:
---------------------------
No idea

Logs/Screenshots:
----------------------------
[//]: # (Please refer to https://github.com/MultiMC/MultiMC5/wiki/Log-Upload for instructions on how to attach your logs.)

No logs.  This happens before running an instance and after.

Additional Info:
---------------------------

Daeruin:
This is happening to me, as well. Choose Minecraft Folder, Config Folder, or Instance Folder from the side menu or the context menu, and MultiMC freezes. It gets as far as giving Finder focus, but MultiMC freezes in the background and has to be Force Quit. If I navigate to the instance folder manually, it appears to have no folders at all. And yet I can still run and play the instance.

peterix:
Yay more macOS bugs.

Wtoll:
This is still happening to me, it's really kind of annoying

45gfg9:
Still today.
macOS 10.15.4
MultiMC 0.6.11
Java 1.8.0_241

Ghalid:
Is this related to #2529? It seems similar to my report, except I also see the freezes from in-game.

[...]

PR #3081, opened 2020-04-06
Stop application freezes on macOS by moving data to MultiMC.app/Data

Ghalid:
macOS seems to dislike when there is a large amount of non-code items in the `Contents` folder. This causes freezes when clicking on links or buttons leading to external locations from MultiMC (see issue #2529).

A simple solution that maintains portability of MultiMC is to move user data, such as instances, into some folder outside of the `Contents` folder but still inside the application bundle. Here, I've chosen `MultiMC.app/Data`.

This PR will change the locations of the following directories and files from `MultiMC.app/Contents/MacOS` to `MultiMC.app/Data`:

```
mods
accounts
accounts.json
assets
cache
icons
instances
libraries
meta
metacache
MultiMC-0.log
MultiMC-1.log
MultiMC-2.log
MultiMC-3.log
MultiMC-4.log
multimc.cfg
themes
translations
```

To prevent existing users from thinking they've lost their data, MultiMC on macOS will check on startup whether data exists in the old location. If it does, it will be moved to the new location.

This PR may also possibly resolve other issues related to freezes/unresponsiveness on macOS: #2579, #2598, #2784, and #3010. However, as I have not experienced full system freezes caused by MultiMC nor freezes when opening the app, I cannot verify that all of these issues would be resolved, this is simply speculation.

Edit: Seems like it may also resolve long startup times as a result of macOS scanning the app (or whatever it does), so issues like #2473 might also be resolved (I discovered that I experience a long "dock bounce time" on a freshly built or freshly downloaded version of the app when I immediately put a large amount of data somewhere in its `Contents` folder (i.e. the instances) before opening it for the first time)

### DIFF
--- a/application/MultiMC.cpp
+++ b/application/MultiMC.cpp
@@ -276,6 +276,43 @@ MultiMC::MultiMC(int &argc, char **argv) : QApplication(argc, argv)
         return;
     }
 
+#if defined(Q_OS_MAC)
+    // move user data to new location if on macOS and it still exists in Contents/MacOS
+    QDir fi(applicationDirPath());
+    QString originalData = fi.absolutePath();
+    // if the config file exists in Contents/MacOS, then user data is still there and needs to moved
+    if (QFileInfo::exists(FS::PathCombine(originalData, "multimc.cfg")))
+    {
+        qDebug() << "On macOS and found config file in old location, moving user data...";
+        QDir dir;
+        QStringList dataFiles {
+            "*.log", // MultiMC-@.log
+            "accounts.json",
+            "accounts",
+            "assets",
+            "cache",
+            "icons",
+            "instances",
+            "libraries",
+            "meta",
+            "metacache",
+            "mods",
+            "multimc.cfg",
+            "themes",
+            "translations"
+        };
+        QDirIterator files(originalData, dataFiles);
+        while (files.hasNext()) {
+            QString filePath(files.next());
+            QString fileName(files.fileName());
+            if (!dir.rename(filePath, FS::PathCombine(dataPath, fileName)))
+            {
+                qWarning() << "Failed to move " << fileName;
+            }
+        }
+    }
+#endif
+
     /*
      * Establish the mechanism for communication with an already running MultiMC that uses the same data path.
      * If there is one, tell it what the user actually wanted to do and exit.
--- a/application/MultiMC.cpp
+++ b/application/MultiMC.cpp
@@ -235,6 +235,10 @@ MultiMC::MultiMC(int &argc, char **argv) : QApplication(argc, argv)
             xdgDataHome = QDir::homePath() + QLatin1String("/.local/share");
         dataPath = xdgDataHome + "/multimc";
         adjustedBy += "XDG standard " + dataPath;
+#elif defined(Q_OS_MAC)
+        QDir foo(FS::PathCombine(applicationDirPath(), "../../Data"));
+        dataPath = foo.absolutePath();
+        adjustedBy += "Fallback to special Mac location " + dataPath;
 #else
         dataPath = applicationDirPath();
         adjustedBy += "Fallback to binary path " + dataPath;


Chew:
I like this a lot, personally putting app data inside the app seems anti-pattern. Should be in Home folder, or Application Data even (however that's pretty hard to get to if you're new). Turns out you can set where instance/mods/icons folders are in the config, but by default shouldn't it be somewhere safer?

Chew:
Also I recommend also moving `cache` as that was about 1 GB for me.

Ghalid:
> I like this a lot, personally putting app data inside the app seems anti-pattern. Should be in Home folder, or Application Data even (however that's pretty hard to get to if you're new). Turns out you can set where instance/mods/icons folders are in the config, but by default shouldn't it be somewhere safer?

I actually fully agree with you. It also has caused issues before when things such as Homebrew updates overwrote the application, deleting all user data; or just user confusion (#1457).

However, I think that it would be straying a little from the intent of this PR to change the data's location to an area outside of the app bundle. Plus I think MultiMC is currently intended to be a portable app, such that a user can simply put the app on a flash drive or similar and carry around their instances (although, I personally don't use it like that).

On the issue I linked above, there's the [following comment](https://github.com/MultiMC/MultiMC5/issues/1457#issuecomment-181517885):

> Looking at this, it is quite a lot to deal with... I'll see what can be done to make the storage local to the machine. It would likely be a separate download though.
> 
> Let's say the portable download will be as it is now. And the 'system-wide' download will contain a special file that will say that all the data and configuration is system-wide.

------------

> Also I recommend also moving `cache` as that was about 1 GB for me.

Hey, you're right - actually, changing the location of the other folders ended up moving the location as `cache` as well - but I'll (force) push a change to move it from the old location to the new location on startup for new users, too.

Boundarybreaker:
If you want to store app data outside the app, Mac provides the `~/Library/Application Support/` folder for doing just that. It's where Minecraft stores all its profiles and such too.

application/MultiMC.cpp
```diff
@@ -272,6 +276,91 @@ MultiMC::MultiMC(int &argc, char **argv) : QApplication(argc, argv)
...
+        }
+        if (!dir.rename(FS::PathCombine(originalData, "multimc.cfg"), FS::PathCombine(dataPath, "multimc.cfg")))
+        {
+            qWarning() << "Failed to move multimc.cfg";
+        }
+        if (!dir.rename(FS::PathCombine(originalData, "accounts.json"), FS::PathCombine(dataPath, "accounts.json")))
+        {
+            qWarning() << "Failed to move accounts.json";
+        }
+        if (!dir.rename(FS::PathCombine(originalData, "MultiMC-0.log"), FS::PathCombine(dataPath, "MultiMC-0.log")))
```
Janrupf, line 341: Not sure if you really need to move the logs, but if you do, what about iterating the files and moving them, if they end with `.log`. I think that would be more readable

Ghalid: [Done.](https://github.com/MultiMC/MultiMC5/pull/3081/commits/8bb5ba76f114b69bfe349160929ef793f9151734) I also decided to iterate through all the data files to move, not just the logs.

peterix:
Hmm. I'll have to test this manually first.

scribbler957:
As a user, I just made a MultiMC folder in my home directory and move the MMC folders there. I don't think that ~/Library/Application Support is really a good choice because it hides user data, in a bloated, hidden directory, and hinders access to the folders when making changes. It also makes the files harder to backup or mirror.

If portability is an issue, maybe there can be an option in settings to save data in the app bundle, or even to the app directory which should work when the app's not in the Applications folder.

Chew:
Do note Minecraft itself stores data there, so it shouldn't be unheard of. You rarely need to access there anyway (if you do, just click open instance folder which will take you there)

scribbler957:
Oh, I know MC stores data in the app bundle. It's bad practice, but I'd be surprised if most programs didn't. Right now there's a few open issues with opening folders and links from MultiMC on the Mac.

Ghalid:
> MC stores data in the app bundle.

Do you mean in `~/Library/Application Support`? The vanilla launcher doesn't keep data in the app bundle (most programs don't), but instead in the aforementioned folder, which Apple has placed there specifically for this purpose.

> Right now there's a few open issues with opening folders and links from MultiMC on the Mac.

Those issues are exactly what I'm aiming to fix in this PR, and they've worked in my own experience :P

scribbler957:
I've seen vanilla leave stuff in there before, but I haven't looked in quite a while and could even be mistaken. I know they're using Application Support now. Anyway, I've been moving things like Steam out of there, but it's still a few hundred GB. Apple would have everything hidden in Library or dumped in a big pile in Documents. Personally, I don't think it's a good plan on their part.
On the other hand, all of you have my admiration and gratitude for the hard work and skill you've put into this project and your generosity in not only sharing it with us, but in taking the time to help deal with these issues.

Ghalid:
> Apple would have everything hidden in Library or dumped in a big pile in Documents. Personally, I don't think it's a good plan on their part.

To be fair, to my knowledge Windows' equivalent to Application Support is basically `%appdata%`, which is kind of hidden in a similar manner (correct me if I'm wrong, I don't use Windows very often). But yeah, I see what you mean.

scribbler957:
You're not wrong, except Application Support is more like App Data, the registry, every program's preferences, and random chunks of the various legacy program directories. To be fair, Apple did have guidelines in place for how these things were supposed to be managed but...

kb-1000:
MultiMC is supposed to be portable by default. That applies to both the windows and linux versions by default as well, they save their settings in the directory where the binaries are. Here they're stored inside the app bundle. However... there's such a "system-wide" version on Linux as well, storing the data in an XDG directory (not sure which one, its .local/share i think), so adding a build define for that could possibly be an option (I'm not the one to decide on that though)

Chew:
Some apps (notably Adobe apps) put a folder in the Applications folder, and inside that folder have the app and also documents. This could be an alternative as well

Ghalid:
> Some apps (notably Adobe apps) put a folder in the Applications folder, and inside that folder have the app and also documents. This could be an alternative as well

I don't think this can be done with MultiMC as easily, as Adobe apps have an installer that creates these folders for you, but MultiMC does not use an installer, so users would have to create such a folder themselves as opposed to MultiMC just selecting a sane directory for data on its own

EthanBar:
Can confirm this fixes the freezing issues on Big Sur. Any progress on this being merged?

Zetabite:
Probably not, as its conflicts with MMC's portability

Chew:
> Probably not, as its conflicts with MMC's portability

No, it won't? It's still inside the .app folder/file from what I understand. It's still one thing to bring over.

peterix:
What it conflicts with is people switching to older versions while this is in place.

You would end up with two copies of the data inside the .app, because the older version on `stable` will not understand this.

peterix:
And that is also a reason why I'm reluctant to merge it.

peterix:
1. Someone going from `develop` to `stable` will think their data is gone.
2. This may lead to confusion and even more data loss because they start messing with it.

We need to warn people about this when it gets merged.

kb-1000:
You could do it in two stages: one version that can read both, and after that has been pushed to stable, this

peterix:
Alternatively:
- Have it use the new location by default if no data exists yet.
- If old data exists, warn people about the implications of migrating it and give them the option to do it. If they say no, just continue using the current data root.

Ghalid:
> What it conflicts with is people switching to older versions while this is in place.
> 
> You would end up with two copies of the data inside the .app, because the older version on `stable` will not understand this.

That’s an important concern, I didn’t think about that.

I’ve been busy, so it’ll likely be a while before I can get back to working on this PR, though. Adding some documentation about this workaround (putting instances and other large folders somewhere other than the `Contents` folder in MultiMC’s settings) for people experiencing this issue may be a good stopgap until then.

hamarb123:
Please see #3919 for an updated version of this PR which asks the user with a dialogue and lets them move it later if they say no.

Janrupf:
Superseded by #3919

PR #3092, opened 2020-04-13
Helper script to copy over a full instance to other hosts

schuellerf:
Any direction works. From your host or to your host or even between two other hosts


peterix:
I'm not really sure where this fits in. Feels like you could just use `rsync` and not add this to the project.

PR #3101, opened 2020-04-18
Fix compiling of api/logic/Version.cpp

johnnyapol:
When building multimc on my machine (Arch Linux, clang 10 using multimc-git via AUR), the build fails with the following error:

image

This simply adds a reference as described in the warning turned error.

### DIFF
--- a/api/logic/Version.cpp
+++ b/api/logic/Version.cpp
@@ -78,7 +78,7 @@ void Version::parse()
     // FIXME: this is bad. versions can contain a lot more separators...
     QStringList parts = m_string.split('.');
 
-    for (const auto part : parts)
+    for (const auto &part : parts)
     {
         m_sections.append(Section(part));
     }


johnnyapol:
These are the custom compiler flags I use for packages to help reproduce:
```
#-- Compiler and Linker Flags
CPPFLAGS="-D_FORTIFY_SOURCE=2"
CFLAGS="-march=x86-64 -mtune=native -O3 -pipe -fno-plt"
CXXFLAGS="-march=x86-64 -mtune=native -O3 -pipe -fno-plt"
LDFLAGS="-Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now"
#-- Make Flags: change this for DistCC/SMP systems
MAKEFLAGS="-j8"
```

kb-1000:
Is there perhaps a -Werror somewhere too?

johnnyapol:
Not that I'm setting. The root CMakeLists.txt has it (https://github.com/MultiMC/MultiMC5/blob/develop/CMakeLists.txt#L35) and I assume that applies to the sub-projects (I'm not super familiar with cmake).

peterix:
Yes. `-Werror` should be mandatory in all C and C++ projects.

peterix:
Please fix the commit message.
As it isn't connected to any github issue, it should start with `NOISSUE `

johnnyapol:
Changed commit message to start with NOISSUE.

PR #3185, opened 2020-06-05
Remove broken Mojang services status

kb-1000:
This can be reverted once the status service is fixed.


peterix:
Is it really this broken, and needs to be removed?

kb-1000:
Well, I would answer that with yes. However I can only really see that the Session indicator is broken and not if the other ones are as well. As you work at Mojang, do _you_ know that?

OpenDeveloper16:
Yea the Session part of the status toolbar is broken... I can't figure out how to upload a picture though

OpenDeveloper16:
Annotation 2020-08-13 092318

OpenDeveloper16:
There we go

Zetabite:
I agree that it should be removed as of now, as most people ask if that has to do with the problem they are having or write this is a problem to begin with

Janrupf:
Can be closed now, superseded by #3524

kb-1000:
Since now the Mojang status check API has been removed entirely, I'm reopening this PR.

kb-1000:
... if GitHub allowed me to? Apparently, it doesn't like that I force pushed to the branch.

PR #3187, opened 2020-06-05
Add basic RPM package for RHEL-based Linux distros

ProjectSynchro:
Title.

I saw another pull request of this nature and figured I'd create one that doesn't require any source binaries to be built.

This is literally a carbon copy of the Debian package but for Fedora, should work for x86_64 and i386.

### DIFF
--- a/application/package/fedora/multimc/multimc.spec
+++ b/application/package/fedora/multimc/multimc.spec
@@ -1,6 +1,6 @@
 ##Init variables
 
-%global packageVer 1.0
+%global packageVer 1.1
 %global _optdir /opt
 
 ## Package info declaration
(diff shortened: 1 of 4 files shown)


peterix:
I think the package could actually be `noarch`, if that's still an RPM packaging concept.

ProjectSynchro:
This is done, however it should only install on ix86 and x86_64 systems as ARM builds aren't available and it would be a bit weird to allow systems to install a package that doesn't work.

peterix:
Nice :)

dhalucario:
Is there a Repository available for this?

ProjectSynchro:
@dhalucario I have a repository hosted on the Fedora Copr
Should work on all supported CentOS, Fedora, and SUSE releases.

https://copr.fedorainfracloud.org/coprs/jackgreiner/multimc/

ifurther:
Dose it has CentOS 7 package ?

ProjectSynchro:
Yes, there is a package for CentOS 7.

ifurther:
OK, thank you

Chronomly:
Any timeline on this being merged?

Krutonium:
https://github.com/MultiMC/MultiMC5/commit/58c2228247c633805cdbe02ff97eb8cb6f3b3af7

Well they did this instead for some reason

ProjectSynchro:
Well leave it to others to reinvent the wheel I suppose, closing this as it appears that clicking the merge button was harder than hastily rewriting this spec.

PR #3195, opened 2020-06-08
Technic pack import

kb-1000:


kb-1000:
Ready for review

application/dialogs/NewInstanceDialog.cpp
```diff
@@ -122,13 +124,15 @@ QList<BasePage *> NewInstanceDialog::getPages()
...
     twitchPage = new TwitchPage(this);
+    auto technicPage = new TechnicPage(this);
     return
     {
         new VanillaPage(this),
         importPage,
         new FtbPage(this),
         new LegacyFTB::Page(this),
-        twitchPage
+        twitchPage,
```
jamierocks, line 134: I've been keeping to alphabetical order for the modpack sources, ideally Technic would come before Twitch.

jamierocks:
I tried to install Tekkit Classic with this and got this error Screenshot_20200831_150910
```
   20.331 C Exception: "Document: Error parsing JSON: illegal value"
   20.331 C Task "Technic::SolderPackInstallTask(0x2e3d4d0)" failed:  "Could not understand \"version.json\":\nDocument: Error parsing JSON: illegal value"
   20.344 C Task "InstanceStaging(0x2e1f9a0)" failed:  "Could not understand \"version.json\":\nDocument: Error parsing JSON: illegal value"
```

kb-1000:
> This one is because your handling of JSON fields is using require where you should be using ensure.

Nope. There is currently no way to import a pack that doesn't contain inheritsFrom. What would you use as the default value for it?

jamierocks:
> > This one is because your handling of JSON fields is using require where you should be using ensure.
> 
> Nope. There is currently no way to import a pack that doesn't contain inheritsFrom. What would you use as the default value for it?

I've just used my tool to see the files Tekkit has, and it doesn't even seem to have a version.json?

```
╭─jmansfield@claude ~/proj/jamierocks/mcinstall/run/tekkit ‹master*› 
╰─$ ls -l bin
total 1928
-rw-rw-r--. 1 jmansfield jmansfield 1972443 Aug 31 16:30 modpack.jar
```

jamierocks:
Oh! I see, of course - `version.json` from the Forge jar.

Yeah, default to the Minecraft version specified by the API.

kb-1000:
Well, how would I pass the version from the API into the TechnicPackProcessor? Remember, I'm also allowing the import of single ZIP files independent of the Technic search page, and I do not want to add Technic specific InstanceImportTask input args either.

jamierocks:
> I'm also allowing the import of single ZIP files independent of the Technic search page

That's functionality not worth having if we can't support some of the largest packs on the platform.

kb-1000:
Okay, so you are suggesting that I should remove it in favor of a new InstanceTask, correct?

jamierocks:
I am.

kb-1000:
I will not drop it, but I think I will indeed change it to use a dedicated import Task for imports from the search page. As long as a pack _can_  be supported with independent single file import, I see no reason for not doing it.

kb-1000:
Ready for review again

peterix:
I am pulling the trigger. Let's expose it to users.

JerryWester:
Neat, I was considering asking if this was still being worked on. Thanks for implementing this!

Issue #2133, opened 2018-02-06
Idea - Add Raspbian Support


afyber:
Would it be possible to add support for Raspbian? When I try to run the generic 32-bit it says:
```
MultiMC Dir: /home/pi/MultiMC
No missing dependencies found.
./MultiMC: line 38: /home/pi/MultiMC/bin/MultiMC: cannot execute binary file: Exec format error
```

DestyNova:
I'd like to help with this and have access to an RPi4. Is there an easy way to get started with this? It's been a long time since I wrote C++ code :grinning:

DestyNova:
Ok! Got it to build using the following commands:
```
mkdir build && cd build
cmake -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang ..
make -j3
```

At first I had a problem with the Qt UI being waaaay bigger than my screen, perhaps to do with some sort of pixel scaling issue in my config. I fixed it by applying this environment variable:

```
QT_AUTO_SCREEN_SCALE_FACTOR=0
```
I've added this line to /etc/environment which I presume will sort the issue for a few other apps too (VLC and Xmoto both have been looking weird on this Pi for a while).

Then I tried to run MultiMC with 1.15.2 and, as expected, ran into a problem with lwjgl:
```
org.multimc.EntryPoint.main(EntryPoint.java:34)
Caused by: java.lang.RuntimeException: java.lang.UnsatisfiedLinkError: Failed to locate library: liblwjgl32.so
        at cxm.<clinit>(SourceFile:38)
        ... 11 more
Caused by: java.lang.UnsatisfiedLinkError: Failed to locate library: liblwjgl32.so
```

Will try to make more progress tomorrow.

peterix:
Welp, keep me posted.

https://github.com/MultiMC/MultiMC5/issues/2133#issuecomment-363351735 is still completely valid, but if you can get this to the point where people can build and use it on their own, there's a chance.

JJTech0130:
I have it built and provide a patcher for it... it only works when the pi is in ARM64 mode though...
All you need to do is put the patcher as the pre-launch command. I also have up-to date arm64  builds. Check [here](https://github.com/JJTech0130/MultiMC5/releases) for builds and [here](https://github.com/JJTech0130/MultiMC-Patcher) for the patcher.

JJTech0130:
[...]

PR #3234, opened 2020-07-08
Add support for custom meta URLs

JJTech0130:
Allows a custom meta URL to be specified when building.
Eg. `cmake -D MultiMC_META_URL:STRING="https://meta.multimc.org/v1/"`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -51,6 +51,7 @@ set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Werror=return-type")
 
 ######## Set URLs ########
 set(MultiMC_NEWS_RSS_URL "https://multimc.org/rss.xml" CACHE STRING "URL to fetch MultiMC's news RSS feed from.")
+set(MultiMC_META_URL "https://meta.multimc.org/v1/" CACHE STRING "URL to fetch MultiMC's meta files from.")
 
 ######## Set version numbers ########
 set(MultiMC_VERSION_MAJOR    0)
--- a/api/logic/CMakeLists.txt
+++ b/api/logic/CMakeLists.txt
@@ -3,7 +3,7 @@ project(MultiMC_logic)
 include (UnitTest)
 
 # Add meta url to meta file
-configure_file("meta/BaseEntity.cpp.in" "meta/BaseEntity.cpp")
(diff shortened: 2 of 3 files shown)


peterix:
Why?

JJTech0130:
So that the ARM64/32 builds can use the official code and don't have to use a seprate fork

kb-1000:
Besides that I heavily disagree with this change, just doing configure_file on all files that need custom strings is very bad style IMO.

JJTech0130:
As for the configure_file thing, some of the other ways it might be done are:

1.  Setup a build config file in the logic section like the one used in the application section, or
2. Pass the url from the application section -> some other things -> the meta file (as an argument)

kb-1000:
Or just have it continue being a fork and change your discord and gh bugtracker URL. That would cause WAY less confusion...

JJTech0130:
And ARM support could be added in another way, such as having MultiMC correctly detect ARM architectures and feed different libs (from the same meta repo) to the instance being launched, as currently it just assumes anything other x64 must be intel 32 bit.

kb-1000:
If you want that... then ask mojang to host arm binaries as well. Otherwise doing that would cause MultiMC no longer being able to be considered as "Vanilla".

JJTech0130:
I think the community has already asked for that, and they said “no”. We used to have Minecraft: Pi edition, but the RPi is much more powerful now and that got abandoned. https://github.com/MultiMC/MultiMC5/issues/2133 asked for raspberry pi support, and I don’t have the time or resources to continue the arm fork indefinitely. It was supposed to be a temporary measure until it got added officially.

peterix:
Anyway, this is fairly benign. Let's merge it.

peterix:
Or at least something like it...

peterix:
Ok, looking at this.

I'm just going to take it line by line and make some changes on top, because we already have one configured file, and the URL can be put in that, instead of making more of the codebase configured.

peterix:
OK. This can now be considered as merged. I've redone it with some changes.

kb-1000:
How would you handle support requests from users using unofficial, non-updated meta repos?

peterix:
I would tell them to go away and bother whoever did that.

peterix:
Such a change means they are no longer using MultiMC, but something else. The name has to be changed, as per the license.

kb-1000:
I fully agree with this... however, this is even a step in the other direction. This makes it really simple to change the meta URL, which now doesn't even require a fork anymore. However to change the name, GitHub issues URL and Discord invite, a fork would definitely be needed.

peterix:
Hmm. You are right.

I will reconsider this.

PR #3237, opened 2020-07-08
It's 2020, right?

JJTech0130:
Pretty sure this should be &copy; 2013-2020


peterix:
Well, probably. It would be easier to mass-replace the year in everything than to edit each individual file in an editor.

JJTech0130:
Yeah...  That's why I stopped 😄.  Someone who knows how to use sed and find properly should do it

PR #3264, opened 2020-07-29
Restructure ubuntu package build rules to match modern packaging

petterreinholdtsen:
Change how the ubuntu package is built to store build rules in the
debian/ directory similar to how Debian and Ubuntu build their packages,
and bring the build setup closer to official packaging, including
a changelog and storing the package files in /usr/.


petterreinholdtsen:
It is unclear to me which version number is most appropriate to use in debian/changelog.  'git tag' show no 1.4 version.

Zetabite:
You know though why MMC is stored in ~/.local/share/multimc?

petterreinholdtsen:
[Adrian]
> You know though why MMC is stored in ~/.local/share/multimc?

My guess is because MultiMC is storing everything below the location of
the binary, and thus do not work well if stored anywhere outside a
single users home directory.

It would be better if it was separating binary location from data
storage.  That would require some code changes.

-- 
Happy hacking
Petter Reinholdtsen

Zetabite:
Its intetional to keep data and binary at same place, to have it portable and no need for code changes depending on what version you download

petterreinholdtsen:
[Adrian]
> Its intetional to keep data and binary at same place, to have it
> portable and no need for code changes depending on what version you
> download

Not sure I understand what you are saying here.  Are you saying that it
is not safe to upgrade the MultiMC binary without throwing away the data
in ~/.local/share/multimc/?

Anyway, I did not try to modify the behavior of the startup script, so
it is still downloading a binary from the Internet and storing it in the
users home directory, without any signature checking of the downloaded
binary.
-- 
Happy hacking
Petter Reinholdtsen

Zetabite:
> Not sure I understand what you are saying here. Are you saying that it is not safe to upgrade the MultiMC binary without throwing away the data in ~/.local/share/multimc/?

Its safe. If you use the normal mmc install via .deb, it will just replace the binary data, but will keep the instances in place. Doing it like you suggest, would store it in a place where the user doesnt have the full access, thats especially nasty for texture pack stuff

petterreinholdtsen:
[Adrian]
> Its safe. If you use the normal mmc install via .deb, it will just
> replace the binary data, but will keep the instances in place.

Aha.  What exactly do you mean by 'binary data' here?  The .so library
files?

Btw, where are the build rules for 'the normal mmc install via .dev'?
The way you phrase it make me suggest the 'launcher' ubuntu .deb is
different from 'the normal mmc install'.

> Doing it like you suggest, would store it in a place where the user
> doesnt have the full access, thats especially nasty for texture pack
> stuff

Not sure what you mean by 'like you suggest' here, as I have not
suggested to change anything affecting user access, as this pull request
move the startup script and icon from /opt/ to /usr/, both out of reach
for a normal user.

-- 
Happy hacking
Petter Reinholdtsen

Zetabite:
Yeah, i was confused, I apologise, I thought you tried to move the ~/.local/share/multimc data to /usr/bin which would have been total BS, but seeing as you want to move things from /opt/ to /usr/, thats fine, though i dont know if that conflicts with older debian versions. I can only tell for arch, where I install it via AUR, where PKGBUILD states everything where it needs to go

petterreinholdtsen:
[Adrian]
> Yeah, i was confused , I apologise, I thought you tried to move the
> ~/.local/share/multimc data to /usr/bin which would have been total
> BS,

Aha.

> but seeing as you want to move things from /opt/ to /usr/, thats fine,
> though i dont know if that conflicts with older debian versions.

It will not conflict with older Debian versions.  The /usr/ namespace is
ment for files originating from .deb packages, and as far as I can tell,
no other packages in Debian uses the multimc name.

-- 
Happy hacking
Petter Reinholdtsen

Zetabite:
> It will not conflict with older Debian versions. The /usr/ namespace is ment for files originating from .deb packages, and as far as I can tell, no other packages in Debian uses the multimc name.

Yeah most likely not. I thought that there was a reason why it was /opt/ for debian

petterreinholdtsen:
Right.  Good to have that cleared away.  I hope the changes can be considered for inclusion soon, to make life easier for the future .deb maintainers. :)

petterreinholdtsen:
Do you need me to clean up the conflict or can you merge the patch as it is?

peterix:
Hm.

This conflicts quite heavily now.
I also don't understand what the value of doing this is... I can produce prefectly fine deb packages with the existing setup. What does this do, aside from rearrange things and require more dependencies to build them?

peterix:
Btw, I build these on Arch, and only test on Ubuntu. I'm not sure if the build deps are even going to do anything.

petterreinholdtsen:
[Petr Mrázek]
> This conflicts quite heavily now.

Aha.  Did not see any conflicts myself, but rebased it on the 'develop'
branch and pushed an update in the hope to get rid of any conflicts.

> I also don't understand what the value of doing this is... I can
> produce prefectly fine deb packages with the existing setup. What does
> this do, aside from rearrange things and require more dependencies to
> build them?

I guess our definitions of 'perfectly fine' differ.

The patch bring the build rules in line with best practice for Debian
source packages on Debian and Ubuntun, instead of synthesising the
package structure manually.  This make it possible to create a
rebuildable .dsc (Debian Source package).  For example, it places files
in debian/ and leave it to the build system to combine the deb content
during build.  The #DEBHELPER# marker and the debian/compat will ensure
any common fragments are inserted into the appropriate scripts.  Using
'dh' to build ensure all the common build flags for debian packages are
automatically applied.  Such things are best practice in Debian these
days.

-- 
Happy hacking
Petter Reinholdtsen

peterix:
I am not following any sort of best practices here... and my definition of them would differ. MultiMC is not a normal linux packageable thing.

peterix:
And I build these on Arch, not sure where would I even get these devscripts...

PR #3268, opened 2020-08-03
"Fixed" broken sessionserver status

Summerfeeling:
Mojang seems to be sending 404 when accessing https://sessionserver.mojang.com/ but looking up a profile works.
So this might be a dirty, but working "solution".


kb-1000:
what
no
that link is only used in the UI
MMC does _not_ check status itself, mojang's status service does, and it sends false data

Summerfeeling:
Ouh. I actually didn't get a look at the data source. I thought it was being used to check the status. Nohting said! I'm sorry 😅

peterix:
Yeah ... we rely on the mojang service status API endpoint to give us something. It hasn't done that for the session service for a while now...

peterix:
It may be necessary to replace this stuff entirely with our own service uptime checker running on a server somewhere.

PR #3299, opened 2020-08-19
Update instance icons, using the newer minecraft textures

QazCetelic:


UnlimitedCookies:
Some prefer the old icons, how about making both an option?

peterix:
I would honestly prefer replacing them with something less lazy. This has been around for ... 8 years?
Maybe redo all the icons in the same style.

peterix:
But you can replace them yourself right now if you like these more... Just put them in the icons folder.

QazCetelic:
> I would honestly prefer replacing them with something less lazy. This has been around for ... 8 years?
> Maybe redo all the icons in the same style.

What about a 3d render of the blocks?

Issue #3095, opened 2020-04-15
Support new FTB API


jamierocks:
## Introduction

Creeperhost have just [released the new Feed The Beast pack installer](https://forum.feed-the-beast.com/threads/ftb-update-new-launcher-ftb-omnia-released.302085/#post-2020272) (they call FTB App). It uses a totally revamped system than the older FTB Launcher, involving a mostly nicely designed API - documented [here](https://modpacksch.docs.apiary.io/).

## Potential Obstacles

There are many ways you could go about implementing support, however note the following:
1. There is currently no way to list all packs
2. There is currently no pagination for any routes
3. All routes are seemingly limited to 50 results

The last 2 points don't currently bear much weight as there aren't 50 packs in the system, so no pagination is required - though the 1st point holds.

## Potential Solutions

I see a few potential solutions here, but I'll list 2 that I think would be the most practical:

1. Only display promoted/latest/popular packs as per the routes available from the API, similar to CreeperHost's FTB App.
2. Crawl their API for packs and add such data to meta (I have done a [similar solution](https://github.com/jamierocks/neptune-ftb-meta) for my own tooling).
    - Alternatively, crawling of their API could be done in MultiMC.

RecurringRaccoon:
Yes please! Lots of FTB packs on CurseForge are already outdated since they've been pushing updates exclusively to FTB App. Seems like they're about to take them down, too. Supporting this API is necessary for Feed The Beast players, even if it's only search and install.

peterix:
Merged into develop

PR #3300, opened 2020-08-19
FTB Support (modpacks.ch)

jamierocks:
Screenshot_20200819_232215

## TODO:

- [x] check `clientonly` and `serveronly` flags
- [x] Cleanup includes

### DIFF
--- a/api/logic/modplatform/modpacksch/PackInstallTask.cpp
+++ b/api/logic/modplatform/modpacksch/PackInstallTask.cpp
@@ -1,11 +1,17 @@
+#include <BuildConfig.h>
 #include "PackInstallTask.h"
+
+#include "FileSystem.h"
+#include "settings/INISettingsObject.h"
+#include "minecraft/MinecraftInstance.h"
+#include "minecraft/PackProfile.h"
 
 namespace ModpacksCH {
 
 PackInstallTask::PackInstallTask(Modpack pack, QString version)
 {
     m_pack = pack;
-    m_version = version;
+    m_version_name = version;
 }
 
 bool PackInstallTask::abort()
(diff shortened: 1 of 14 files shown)


jamierocks:
Resolves GH-3095.

application/pages/modplatform/ftb/FtbModel.cpp
```diff
@@ -0,0 +1,250 @@
...
+    if(parse_error.error != QJsonParseError::NoError) {
+        qWarning() << "Error while parsing JSON response from FTB at " << parse_error.offset << " reason: " << parse_error.errorString();
+        qWarning() << response;
+        return;
+    }
+
+    auto obj = doc.object();
+
+    ModpacksCH::Modpack pack;
+    ModpacksCH::loadModpack(pack, obj);
```
peterix, line 184: Catch parsing errors here. Basically all of those `require*` JSON calls will throw if there's any problem. Uncaught exceptions lead to a crash.

api/logic/modplatform/modpacksch/PackInstallTask.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if(parse_error.error != QJsonParseError::NoError) {
+        qWarning() << "Error while parsing JSON response from FTB at " << parse_error.offset << " reason: " << parse_error.errorString();
+        qWarning() << response;
+        return;
+    }
+
+    auto obj = doc.object();
+
+    ModpacksCH::Version version;
+    ModpacksCH::loadVersion(version, obj);
```
peterix, line 68: Same here, this needs to be guarded against errors.

peterix (changes requested):
Some small changes needed to avoid unhandled exceptions.

api/logic/modplatform/modpacksch/PackInstallTask.cpp
```diff
@@ -0,0 +1,150 @@
...
+    MinecraftInstance instance(m_globalSettings, instanceSettings, m_stagingPath);
+    auto components = instance.getPackProfile();
+    components->buildingFromScratch();
+
+    for(auto target : m_version.targets) {
+        if(target.type == "game" && target.name == "minecraft") {
+            components->setComponentVersion("net.minecraft", target.version, true);
+            continue;
+        }
+    }
```
jamierocks, line 106: Should this error out if the Minecraft version isn't found?

application/pages/modplatform/ftb/FtbModel.cpp
```diff
@@ -0,0 +1,259 @@
...
+    {
+        QVariant v;
+        v.setValue(pack);
+        return v;
+    }
+
+    return QVariant();
+}
+
+void ListModel::performPaginatedSearch()
```
jamierocks, line 71: Should rename this, as the search isn't paginated.

To add onto this, the API merely returns a list of pack IDs (that need to then be each queried) - we can implement pagination by batching this process, querying for X amount at a time (from the list of IDs we currently have).

If I get chance, I'll look into how pagination with the ListModel works, else this may form a future pr.

peterix:
Squashed, improved and merged into develop

PR #3313, opened 2020-08-24
Enable -Wmissing-declarations warnings

AbigailBuccaneer:
This warns whenever a function (other than a `static` or `inline` one) is defined without a separate declaration. This ensures that file-local functions get marked as `static` as they should be. This helps the compiler recognize dead code more effectively (the `-Wunused-function` warning, enabled by `-Wall`, can only recognise unused static functions.) It also helps the optimizer inline more aggressively (as it knows every call site of a static function).


peterix:
I don't think we need this.

Issue #405, opened 2014-07-28
Ability to manage AT Launcher's modpacks right from MultiMC


KamranMackey:
I would find it really useful if we could manage AT Launcher's mod packs since you have FTB integration, so why not add AT Launcher integration as well?

Aphenon:
Was this feature dropped? i would love to have this.

RyanTheAllmighty:
I'm sad this never made it in. The fact it's still open provides hope, but only very little since it's been left here.

Any update on this or how I can help in anyway make this happen?

peterix:
Yep. We don't have enough manpower.

Currently, all my effort goes into this:
https://github.com/MultiMC/MultiMC5/commits/feature/remote_version_files

Basically making a new instance format that can handle modpacks and direct resource downloads. Kinda like QuickMods should have been, only built from bottom up, not the other way.

peterix:
Long time no see! I'm moving all the ideas to workflowy - a place better suited for the kind of structured organization and planning needed to actually manage hundreds of ideas.

The feature will most likely take the form of direct support - and import from the platform. Not tracking or launching of existing ATL instances.

So, this has been [moved to workflowy (Ideas -> Packs).](https://workflowy.com/s/2EyDMcp7CU)

peterix:
So, I'm looking at this, and I don't think this can ever go forward.

1. I don't want to reimplement other launcher's logic just to be able to launch some arbitrary instances from them. It would tie MultiMC to their internals. It did that with FTB, and I'm not a fan of the mess it causes in the code. The launchers go in different directions, and the differences are often irreconcilable.
2. Import from platform is not possible.

I might do this in the 'direct import from platform' way on a private branch just for the heck of it... But it cannot be deployed.

PR #3314, opened 2020-08-24
GH-405 ATLauncher Support

jamierocks:
Screenshot_20200825_134017

This may not ever be complete enough to reasonably be merged - there are a few things that ATLauncher does that may not be able to work (well) under MultiMC.

- Browser downloads (where the user is instructed to manually download a mod)
- Detecting loaders (Forge, LiteLoader) in packs using the legacy mechanism (challenge shared in part with the work being done for Technic support)
- ~~Pack icons (ATLauncher uses rectangular pack icons)~~

The following is tasks that ought to be implemented for supporting installing ATLauncher packs:

- Support selecting optional mods (likely with a GUI the same as ATLauncher's own)
- ~~Sort packs by game version, alphabetically, and popularity (think MMC's legacy FTB support, and the display in the ATLauncher)~~
- Show private packs if any account has access to it

Low level tasks that still need to be done:

- Support the full range of [mod types](https://wiki.atlauncher.com/mod_types)
- ~~Download and extract the pack's configuration files (this depends on an "extracting download task" that's a shared goal with Technic support)~~
- ~~Get the ATLauncher icon into the theme~~
- Support using latest, recommended, and choose [loader](https://wiki.atlauncher.com/xml:loaders) versions
- ~~Preliminary support for packs using the legacy loader mechanisms~~

## Pack Support

| Pack Name | Undetected | Detected | Notes |
| --- | --- | --- | --- |
| Sky Factory 4 - 4.2.2 |  | Minecraft Forge |  |
| Sky Factory - 2.4 |  | Minecraft Forge |  |
| Yogscast Complete Pack - 2.9.3.4-RR-YOGS | Liteloader | Minecraft Forge | |
| Bytesize - 0.1.7 |  | Minecraft Forge |  |
| Ye Olde Times - 0.2 | | Minecraft Forge | Has jarmods (CodeChickenCore, NotEnoughItems) |

---

Fixes GH-405.


wlhlm:
First of all, thanks for your work on integrating third-party pack platforms into MMC ❤️ 

> there are a few things that ATLauncher does that may not be able to work (well) under MultiMC.

How about adding a way to import ATLauncher packs from disk, similar to how the old FTB support used to work? This might simplify the implementation on the MMC side. I see that some of the yet-to-be-implemented ascpects are also helpful for Technic integration, but for me at least, the strengths of MMC are launching and running minecraft instances, not browsing/installing modpacks and I'm worried that these (deeper) integrations will be harder to maintain in the future.

jamierocks:
> How about adding a way to import ATLauncher packs from disk, similar to how the old FTB support used to work?

The core part of that would be near identical to what's happening here already - we'd just save implementing features like browser downloads, etc. It would be more frustrating too, as ATL has 2 different formats it uses locally.

> I'm worried that these (deeper) integrations will be harder to maintain in the future.

I've intentionally made both my FTB and ATL support work as easy to remove at any point as I can - its all in separate files, and has minimal changes to existing code.

api/logic/modplatform/atlauncher/ATLPackInstallTask.cpp
```diff
@@ -0,0 +1,194 @@
...
+            case DownloadType::Unknown:
+                emitFailed(tr("Unknown download type: ") + mod.download_raw);
+                return;
+        }
+
+        qDebug() << "Will download" << url << "to" << path;
+        auto dl = Net::Download::makeFile(url, path);
+        jobPtr->addNetAction(dl);
+    }
+    components->installJarMods(jarmods);
```
jamierocks, line 169: This doesn't work at the moment, it needs to be run after the jarmods exist - unfortunately my attempt to do
```cpp
    connect(jobPtr.get(), &NetJob::succeeded, this, [&]()
    {
        jobPtr.reset();
        components->installJarMods(jarmods);
        instanceSettings->resumeSave();
        emitSucceeded();
    });
```
was unsuccessful - I welcome input on how to handle this *nicely*.

kb-1000:
@jamierocks I see an issue with that:
You use these badges or whatever to call them as logos. That just won't look well later.

api/logic/modplatform/atlauncher/ATLPackInstallTask.cpp
kb-1000: Try getComponents()->saveNow() after installing the mods.

jamierocks:
Well its a choice between using the pack image or using nothing, and it looks fine IMO.

Screenshot_20200825_224855

Zetabite:
i think its fine

jamierocks:
Okay, we're getting somewhere - Sky Factory 4 (a recent pack using modern ATLauncher concepts) is now installable :D 



(screenshot is having connected to a SF4 server)

api/logic/modplatform/atlauncher/ATLPackManifest.cpp
```diff
@@ -0,0 +1,136 @@
...
+    p.url = ele.attribute("url");
+    p.file = ele.attribute("file");
+    p.md5 = ele.attribute("md5");
+
+    p.download_raw = ele.attribute("download");
+    p.download = parseDownloadType(p.download_raw);
+
+    p.type_raw = ele.attribute("type");
+    p.type = parseModType(p.type_raw);
+    p.extractTo_raw = ele.attribute("extractto");
```
RyanTheAllmighty, line 106: Note that here we can have an `extractfolder` property which will extract the value of the `extractfolder` property from the zip file.

For instance say I have a zip with the below structure:

- /
  - somemod/
    - mod-folder/
      - modFile.txt

By not specifying `extractfolder` and if I had `extractto="root"` then I'd have a folder in my root directory called `somemod`. But by providing `extractfolder="somemod%s%sub-folder"` then the `modFile.txt` would end up in the root directory.

Note the `%s%` is used to denote a file path separator

This was mainly used for Millenaire back in the day, but is still a valid option to have.

RyanTheAllmighty:
Some handy (or not) links:

- https://wiki.atlauncher.com/mod_types (wiki information on different mod types and their values/users)
- https://cdn.atlcdn.net/ATLauncher+XML+and+API+Documentation.pdf (older documentation on all the xml options, this generally has more information)
- https://cdn.atlcdn.net/ATLauncher.xsd (schema used to validate XML files)

api/logic/modplatform/atlauncher/ATLPackManifest.cpp
```diff
@@ -0,0 +1,136 @@
...
+    else if(rawType == QString("plugins")) {
+        return ATLauncher::ModType::Plugins;
+    }
+    else if(rawType == QString("extract")) {
+        return ATLauncher::ModType::Extract;
+    }
+    else if(rawType == QString("decomp")) {
+        return ATLauncher::ModType::Decomp;
+    }
+    else if(rawType == QString("resourcepack")) {
```
RyanTheAllmighty, line 62: Note there is also "texturepack" for older packs pre resourcepack directory naming

api/logic/modplatform/atlauncher/ATLPackInstallTask.cpp
```diff
@@ -0,0 +1,416 @@
...
+
+bool PackInstallTask::abort()
+{
+    return true;
+}
+
+void PackInstallTask::executeTask()
+{
+    auto *netJob = new NetJob("ATLauncher::VersionFetch");
+    auto searchUrl = QString(BuildConfig.ATL_DOWNLOAD_SERVER + "packs/%1/versions/%2/Configs.xml")
```
RyanTheAllmighty, line 34: Any reason you're using the XML and not the json? We do not use the XML and do a lot of transformation server side to transform this xml to json.

I'll be removing this xml file from the download server fairly soon

RyanTheAllmighty: Example: https://download.nodecdn.net/containers/atl/packs/SkyFactory4/versions/4.2.2/Configs.json

jamierocks: It doesn't make a lot of difference either way, if the XML is going - I'll move over to the JSON.

My concern was that the JSON seemed to be tied to launcher implementation details, and some data from the XML could be omitted as a result. This was something I meant to discuss with you, just something I forgot about.

api/logic/modplatform/atlauncher/ATLPackManifest.cpp
jamierocks: I figured there would be, I was going by the API docs which doesn't seem to have resource packs and texture packs on there. I'll look at the launcher eventually to make sure all types are handled :)

api/logic/modplatform/atlauncher/ATLPackInstallTask.cpp
RyanTheAllmighty: Yeah I've been meaning to remove the XML for years, and there's a big overhaul of the backend systems coming within the next month, which includes removing these files.

All the same information should be there. There shouldn't be anything missing

api/logic/modplatform/atlauncher/ATLPackManifest.cpp
jamierocks: This should be resolved by https://github.com/MultiMC/MultiMC5/pull/3314/commits/09441ecbdfc67b743eda3444846115e79fa62c2b.

jamierocks:
Before this is merged, https://github.com/MultiMC/MultiMC5/pull/3314/commits/dabb454dd8d53fa5c97cb2fe3c70b812a948a8b1#diff-323da61b4a74d94009e6550ae1103d14 should be done properly.

api/logic/modplatform/atlauncher/ATLPackManifest.cpp
jamierocks: I think I have everything but `millenaire`, `texturepackextract`, and `resourcepackextract` now - what is `millenaire` used for?

api/logic/modplatform/atlauncher/ATLPackInstallTask.cpp
```diff
@@ -108,8 +108,12 @@ QString PackInstallTask::getDirForModType(ModType type, QString raw)
         case ModType::Decomp:
             qWarning() << "Unsupported mod type: " + raw;
             return Q_NULLPTR;
+        case ModType::TexturePack:
+            return "texturepackss";
```
phit, line 112: is this a typo?

jamierocks: it is, good spot!

jamierocks: Resolved by https://github.com/MultiMC/MultiMC5/pull/3314/commits/44a12a9836eb692f2ad23ca03a016bb67b6d616f.

api/logic/modplatform/atlauncher/ATLPackInstallTask.cpp
jamierocks: Resolved by https://github.com/MultiMC/MultiMC5/pull/3314/commits/44a12a9836eb692f2ad23ca03a016bb67b6d616f.

jamierocks:
The libraries for Sky Factory 2.4 are now completely detected (see https://gist.github.com/jamierocks/7a40bf51cd92a7ca6d3a8d4f440cc8ab), will need to test more packs and add further heuristics.

jamierocks:
Sky Factory 2.4 now imports in a components-compliant manor (that is to say using the existing components).

Screenshot_20200831_144323

Further work needs to be done to support detection of the Liteloader version, which can't be done based on the name sadly (`liteloader-1.6.4.jar` is useless to us) - reverse lookup by the hash is likely going to be my approach. When this is implemented, the Yogscast Complete Pack should import compliantly.

jamierocks:
Ye Olde Times, one of the oldest packs on the platform, now installs completely and in a components complaint manor.

Screenshot_20200902_091706

In the future, further name inspection (perhaps simply "Forge") will likely be required - in addition to detecting Forge version based on similar mechanisms to the Technic support for packs where this support cannot be used.

peterix:
Well, this has been stuck in 'Draft' state for a while. Let's see if we can get it in.

peterix:
I just manually rebased, squashed and merged this. Let's see what the users think :)

AverageComet250:
Which release will this be included in?

Zetabite:
Wont be in stable for long time. But already in develop

AverageComet250:
Is that before 0.10 or after?

I'm okay with switching to develop short term until this goes into stable, but I'd rather not be stuck in develop for ages, since I use multimc a lot and I'd rather not find that I suddenly can't launch my Minecraft instances.

kb-1000:
We're not even at 0.7 yet.
Also...
> I'd rather not find that I suddenly can't launch my Minecraft instances.

What?
The format has not changed

AverageComet250:
I know we're still on 0.6.

> The format has not changed

I know it hasn't I meant updating to a commit which won't build.

kb-1000:
If it doesn't build, there is no build to update to.

PR #3360, opened 2020-09-28
Create Windows Installer

njensen20:
A installable version of MultiMC should be made available as an alternative to the portable zip format currently used. Installable programs are more friendly to packages managers such as winget. NSIS was used because it is open source, supports many languages, and is highly flexible. 

The current installer installs to %appdata%/MultiMC, so default instance and asset paths don't need to be changed.

Functioning installer releases are available from my fork as an example of how the installer and uninstaller are currently configured


Heufneutje:
I think you've missed the point of MultiMC being a portable application by design. Not to mention that this is adding maintainability cost. I highly doubt this will be accepted in any way, shape or form.

peterix:
OK.

Assuming I would even want this anywhere near the codebase (and I don't, Windows installers should die in the fires that spawned them, like the One Ring).

1. It does not integrate with the CI and deployment pipelines and introduces more things that need to be hand tweaked for every release. I already 
2. Anything that even resembles the 'click next' experience needs to be yeeted into the abyss.
3. Integration with Windows 'package managers' (and they are so far from package management as they possible could be) is NOT a goal. If MultiMC ends up in them, I will have some stern words with whoever did that and/or send them takedown notices. The only place to get MultiMC is multimc.org or a custom build from source that the direct end user does (I assume they can trust themselves with the integrity of the build).
4. The license part is a copypasta from the real license file (see point 1.).
5. Installing the windows way into APPDATA is pure insanity.
6. This does not even attempt to follow the established commit message format.

I don't need even more maintenance burden and this doesn't add anything worth the drawbacks. If I have to even look at a Windows machine to make a release, there will be no more Windows releases at all.

peterix:
Also, please remove the installers from anywhere they were put on.

peterix:
OR, do not call the result MultiMC.

peterix:
That aside, I think there's a good way to do this, it's just probably not going to look anything like this.

Exploring how it should look is definitely open.

njensen20:
The installer files were removed as requested. 

If you do not want MultiMC in any package manager, you may want to know it has been available through the Chocolately package manager since April (https://chocolatey.org/packages/multimc). It has been downloaded over 1,700 times through Chocolatey.

As far as installing to APPDATA, that is standard practice for programs installing for unelevated users or if the program needs RW access to its own program files. Microsoft Teams and Visual Studio Code install to APPDATA by default and Google Chrome will install to APPDATA if the user isn't elevated.

kb-1000:
Apparently, t downloads the ZIP file from releases: 
```powershell
$ErrorActionPreference = 'Stop';
$url        = 'https://github.com/MultiMC/MultiMC5/releases/download/0.6.11/mmc-stable-win32.zip'
$toolsDir   = "$(Get-ToolsLocation)"

$packageArgs = @{
  packageName   = $env:ChocolateyPackageName
  unzipLocation = $toolsDir
  url           = $url
  softwareName  = 'multimc*'
  checksum      = '007b517746a7659560c9d9abc8a2e42a7497e27337ada4adfad06390446f59bc'
  checksumType  = 'sha256'
}
Install-ChocolateyZipPackage @packageArgs

Write-Host "Creating Start Menu and Desktop shortcuts"
$startmenu = Join-Path $env:programdata "Microsoft\Windows\Start Menu\Programs"
Install-ChocolateyShortcut -ShortcutFilePath $(Join-Path $startmenu "MultiMC.lnk") -TargetPath $(Join-Path $toolsDir "MultiMC\MultiMC.exe")
$desktop = [Environment]::GetFolderPath("Desktop")
Install-ChocolateyShortcut -ShortcutFilePath $(Join-Path $desktop "MultiMC.lnk") -TargetPath $(Join-Path $toolsDir "MultiMC\MultiMC.exe")
```

The uninstall script, however, deletes the instances:
```powershell
Write-Host "Removing Start Menu and Desktop shortcuts"
$StartMenuShortcut = Join-Path $env:programdata "Microsoft\Windows\Start Menu\Programs\MultiMC.lnk"
$DesktopShortcut = Join-Path $([Environment]::GetFolderPath("Desktop")) "MultiMC.lnk"
if (Test-Path $StartMenuShortcut) 
{
  Remove-Item $StartMenuShortcut
}
if (Test-Path $DesktopShortcut)
{
  Remove-Item $DesktopShortcut
}

$multimcdir = Join-Path "$(Get-ToolsLocation)" MultiMC
if(Test-Path $multimcdir)
{
  Write-Host "Your instances will be deleted (if they live in the MultiMC directory: $multimcdir)" -ForegroundColor Black -BackgroundColor Yellow
  Write-Host "Ctrl-C to cancel" -ForegroundColor Black -BackgroundColor Yellow
  timeout 10
  Remove-Item -path "$multimcdir" -recurse
}
```

peterix:
I don't care what the 'standard practice' is. It's awful as usual.

peterix:
If people come in and complain about getting any problems from these things, they will get NO support.

PR #3436, opened 2020-11-12
NOISSUE Corrected the fix for Classic saving, fixes Indev compatibility

Moresteck:
The previous fix made Indev versions unplayable.

### DIFF
--- a/libraries/launcher/net/minecraft/Launcher.java
+++ b/libraries/launcher/net/minecraft/Launcher.java
@@ -143,7 +143,11 @@ public class Launcher extends Applet implements AppletStub
     public URL getDocumentBase()
     {
         try {
-            return new URL("http", "www.minecraft.net", 80, "/game/", null);
+            // Special case only for Classic versions
+            if (wrappedApplet.getClass().getCanonicalName().startsWith("com.mojang")) {
+                return new URL("http", "www.minecraft.net", 80, "/game/", null);
+            }
+            return new URL("http://www.minecraft.net/game/");
         } catch (MalformedURLException e) {
             e.printStackTrace();
         }


peterix:
Can you explain what this does?

It looks like something that should be data-driven from tthe manifests, not hardcoded to look for a random class name.

Moresteck:
It reverts the method's behaviour to the previous one for Indev - 1.5.2 versions.
It wouldn't really matter, if not for Indev's legitimacy check. This check has to be reflected in the applet wrapper in order for Indev to launch.

In(f)dev's legitimacy check (in-20100112 to inf-20100320):
`this.J.getDocumentBase().toString().startsWith("http://www.minecraft.net/") || this.J.getDocumentBase().toString().startsWith("http://minecraft.net/")`

Classic's way of obtaining the ip and port for level saving (before c0.0.16a_02-081026):
`this.getDocumentBase().getHost() + ":" + this.getDocumentBase().getPort()`

Right now, the method returns a URL of value: `http://www.minecraft.net:80/game/`
While it fixes classic's URL detection for online saving, it breaks Indev.
Indev requires the method to return a string which starts with: `http://www.minecraft.net/` (without the `:80` port)

Notch has [acknowledged](https://ia803208.us.archive.org/12/items/Rotab-Minecraft-IRC-logs/%23minecraft.20090607.log) a bug with online saving during Classic mptest and fixed it for c0.0.16a_02-081026:
_(16:59:05) <@Notch> thank you for the nice bug report. =)
(16:59:16) <@Notch> (-1 is a very strange port)_

It is unknown how the bug came to be during mptest, but it's known that the issue didn't exist before.
The exception i've made for versions launched from the `com.mojang` package fixes the issue, and leaves the default option for newer versions (so that Indev works).
Only Classic versions have a classpath starting with `com.mojang`.

PR #3476, opened 2020-12-08
NOISSUE Add metainfo support

joshua-stone:
[Metainfo is an AppStream specification for providing application metadata that can be shared between package formats.](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html)

Key changes:

* Inclusion of multimc.metainfo.xml
* Change in MultiMC5.spec for installing multimc.metainfo.xml

A built RPM should look like:

```
$ rpm -ql ~/rpmbuild/RPMS/x86_64/MultiMC5-*.rpm 
/opt/multimc
/opt/multimc/icon.svg
/opt/multimc/run.sh
/usr/share/applications/multimc.desktop
/usr/share/metainfo/multimc.metainfo.xml
```


kb-1000:
Would this work for the Ubuntu package as well?

peterix:
Indeed. It looks like something both packages could use.

joshua-stone:
@peterix The metainfo file should be usable in both of the supported package formats now.

peterix:
Good stuff :)

joshua-stone:
Yep! It appears that .rpm and .deb are the two primary package formats supported in this repository at the moment. Are there other package formats you want this to be added to?

peterix:
I think I'll have to add it to the AUR package... but that's on me.

peterix (approved):

PR #3478, opened 2020-12-09
NOISSUE Change the install directory from /opt to /usr/share to be inline with Fedora Packaging Guidelines

ProjectSynchro:
As stated [here](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_limited_usage_of_opt_etcopt_and_varopt), it appears that /opt is reserved for special use by some vendors.. I changed some parts of the spec to not directly use anything from the Ubuntu package directory (short of run.sh) ~~and I added a patch for the desktop file in the `application/package/linux` directory.~~

Imo if there is a way to use patches / scripts to bring all the debian related package stuff into that one directory it seems like the most sustainable option (should you want to write other package specs, etc down the line)


ProjectSynchro:
meh, just going to be half baked and not ideal in the end. 
Closing.

PR #3487, opened 2020-12-15
NOISSUE Move fabric up in the modloaders list

PCITechie:
Fabric is now widely regarded as the preferred modloader for Minecraft, also updating much quicker than its older (however still modern-ish) counterpart, Forge, and I think it would be more intuitive for it to be moved up the list.

### DIFF
--- a/application/pages/instance/VersionPage.ui
+++ b/application/pages/instance/VersionPage.ui
@@ -158,20 +158,20 @@
     <string>Revert the selected package to default.</string>
    </property>
   </action>
-  <action name="actionInstall_Forge">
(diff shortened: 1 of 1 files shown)


phit:
> now widely regarded as the preferred modloader

citation needed, just looking at [CurseForge](https://www.curseforge.com/minecraft/modpacks?filter-game-version=2020709689%3A8134&filter-sort=5) only 3 out of the top 20 packs on 1.16.4 are using Fabric

peterix:
Disagree.

This is not a significant enough improvement and seems arbitrary.

PR #3524, opened 2021-01-12
Switch Session Server

TreyRuffy:
sessionserver.mojang.com has been down for a while and it looks like it is no longer in use.
[Link to where Mojang says it is not in use](https://bugs.mojang.com/browse/WEB-3166?focusedCommentId=820167&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-820167)

I have switched the URL to session.minecraft.net

### DIFF
--- a/application/widgets/ServerStatus.cpp
+++ b/application/widgets/ServerStatus.cpp
@@ -65,7 +65,7 @@ ServerStatus::ServerStatus(QWidget *parent, Qt::WindowFlags f) : QWidget(parent,
 
     addStatus("authserver.mojang.com", tr("Auth"));
     addLine();
-    addStatus("sessionserver.mojang.com", tr("Session"));
+    addStatus("session.minecraft.net", tr("Session"));
     addLine();
     addStatus("textures.minecraft.net", tr("Skins"));
     addLine();


Moo-dev:
h

Zetabite:
This isnt better. Now you are leading it to a page where it cant check either

TreyRuffy:
> This isnt better. Now you are leading it to a page where it cant check either

I am not sure I understand. I am still checking [here](https://status.mojang.com/check) and as you can see on the page, `sessionserver.mojang.com` is always red while `session.minecraft.net` will update.

AverageComet250:
`session.minecraft.net` was down the other day, but it's back up now.

Janrupf:
Oh god Mojang... why... but if this works properly, that should be the solution to go.

mdashlw:
This is not true that `sessionserver.mojang.com` is no longer in use, because it is. It's used by all modern servers starting from 1.7-ish. `session.minecraft.net` used to be the old session server for legacy auth, it's probably still up. 
I'm not sure whether `status.mojang.com/check` displays status of legacy `session.minecraft.net` or is messed up and actually means `sessionserver.mojang.com`.
Either way, this change makes sense since `sessionserver` has been red for quite a while.

PR #3564, opened 2021-01-30
Fix OSX build instructions

amaury-d:
Curent OSX instructions relies on GCC48 that has been decomissioned and removed from homebrew. Thus, it doesn't work anymore.

This PR fixes the instructions to compile MultiMC by moving to llvm clang.

https://github.com/amaury-d/MultiMC5/blob/feature/build-on-osx/BUILD.md

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -178,30 +178,34 @@ zlib1.dll
 * install homebrew
 * then:
 
-```
+```sh
+brew install openjdk@8
 brew install qt5
-brew tap homebrew/versions
-brew install gcc48
 brew install cmake
+
+# libc++abi
+git clone https://github.com/llvm/llvm-project.git
+cd llvm-project
+mkdir build-libcxxabi && cd build-libcxxabi
+cmake ../libcxxabi
+make && make install
 ```
 
 ### Build
 
 Pick an installation path - this is where the final `.app` will be constructed when you run `make install`. Supply it as the `CMAKE_INSTALL_PREFIX` argument during CMake configuration.
 
-```
+```sh
 git clone https://github.com/MultiMC/MultiMC5.git
 cd MultiMC5
 git submodule init
 git submodule update
-mkdir build
-cd build
+mkdir build && cd build
+export PATH="/usr/local/opt/openjdk@8/bin:$PATH"
 export CMAKE_PREFIX_PATH=/usr/local/opt/qt5
-export CC=/usr/local/bin/gcc-4.8
-export CXX=/usr/local/bin/g++-4.8
+export CXXFLAGS="-Wno-range-loop-analysis ${CXXFLAGS}"
 cmake .. -DCMAKE_INSTALL_PREFIX:PATH=/Users/YOU/some/path/that/makes/sense/
-make
-make install
+make && make install
 ```
   
 **These build instructions were taken and adapted from https://gist.github.com/number5/7250865 If they don't work for you, let us know on IRC ([Esper/#MultiMC](http://webchat.esper.net/?nick=&channels=MultiMC))!**


AverageComet250:
I can't find the branch your changes are on. Can you please put a link so I can see how build.md will look when pushed. The mobile version only lets me see the actual markdown.

amaury-d:
> I can't find the branch your changes are on. Can you please put a link so I can see how build.md will look when pushed. The mobile app only lets me see the actual markdown.

Sure, the description is edited with the link.

AverageComet250:
I don't think it's necessary to set the compiler to clang when using cmake as I think it uses it by default.

amaury-d:
> I don't think it's necessary to set the compiler to clang when using cmake as I think it uses it by default.

You're right, but without it, it used clang 12 (apple) with `range-loop-analysis` option which caused an error on compilation due to C++17 norm.
I had to switch to homebrew clang which is in version 11 and doesn't cause its issue.

I tried to play with cmake options to disable `range-loop-analysis` but was not successful.

amaury-d:
> I tried to play with cmake options to disable `range-loop-analysis` but was not successful.

I may have found a way to make it work with clang 12 provided by Apple. I will update the pull request.

AverageComet250 (approved):
This is fine now.

peterix:
I'm not sure what you are doing here, but this is definitely not how MultiMC is built on CI.

peterix:
I should really do this from scratch. The instructions were written years ago and were never correct.

amaury-d:
@peterix Closing a PR where people spent time without further explanation is never a good idea and can undermine a whole community motivation.
You could have at least explained "how MultiMC is built on CI"...

Zetabite:
> I should really do this from scratch. The instructions were written years ago and were never correct.

I guess that you are doing something in a way, how it shouldnt be done, but due to the faulty instructions its not clear what to do. so i guess wait till the new instructions are out and revisit your pr to adapt it to the new instructions

peterix:
Well, it goes roughly like this:
- Install XCode
- Install the official build of CMake (https://cmake.org/download/)
- Install JDK 8 (https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html)
- Get Qt 5.6 and install it (https://download.qt.io/new_archive/qt/5.6/5.6.3/)

You can see an example of a build here:

```
cmake \
 -DCMAKE_C_COMPILER=/usr/bin/clang \
 -DCMAKE_CXX_COMPILER=/usr/bin/clang++ \
 -DCMAKE_BUILD_TYPE=Release \
 -DCMAKE_INSTALL_PREFIX:PATH="../dist/" \
 -DCMAKE_PREFIX_PATH="/path/to/Qt5.6/" \
 -DQt5_DIR="/path/to/Qt5.6/" \
 -DMultiMC_LAYOUT=mac-bundle \
 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.7 \
 ..
make install
```

there are some other variables you could set up, but that would only be done for distribution...

At no point does the current MultiMC build environment involve homebrew.

peterix:
I haven't messed with this in a few years, so it's possible I'm missing a step...

AverageComet250:
@peterix those instructions are nearly perfect. Instead of the official jdk you can use openjdk or adopt openjdk.

amaury-d:
@peterix homebrew is indeed not mandatory but was a helper available for osx world instead of installing cmake, jdk8 and qt56 manually.

Concerning libc++abi, it was absent in my Catalina despite a clean xcode install, hence the necessity to install it manually. I would be interested to know if it is the same for you when you'll update those instrutions.

The provided instructions used today in CI are the same that the one provided in the PR minus the supplementary options:
`
 -DCMAKE_BUILD_TYPE=Release \
 -DCMAKE_INSTALL_PREFIX:PATH="../dist/" \
 -DMultiMC_LAYOUT=mac-bundle \
 -DCMAKE_OSX_DEPLOYMENT_TARGET=10.7
`

peterix:
I've put the updated instructions into the BUILD.md

PR #3595, opened 2021-02-10
Add modpack search for ATLauncher packs

Zetabite:
Screenshot_20210210_224902


peterix (changes requested):
Hmmm... a few things.

1. There's no need to re-fetch that json file.
    Specifically, it's possible to have a proxy model in front of the pack model that does filtering based on the full text of the pack dynamically as the user types. With the number of packs, this is entirely possible to do.
2. Please use a prefix to commit messages (NOISSUE or GH-#### if it references a github issue)
3. There are way too many commits. Please rebase and squash.
4. There's a conflict with a change that got merged before this - the UI has changed - and the conflict needs resolving.

Zetabite:
I will fix it

PR #3596, opened 2021-02-10
NOISSUE Add ATLauncher modpack search

Zetabite:
Still needs ui resolve


Zetabite:
The ui still needs some checking, i am not 100% that it will work.

Zetabite:
So far this should work now

jamierocks:
Hey Zetabite :wave: 

I've just opened GH-3597 which achieves the same goal but using the existing filter model and doesn't make further HTTP requests. It's also a fair bit smaller as a result.

Zetabite:
*sigh*

peterix:
Hmm. The other PR got merged.

PR #3597, opened 2021-02-11
NOISSUE Add search to ATLauncher

jamierocks:
Screenshot_20210211_003135


peterix (commented):
I don't think we need the search button with this. Just update as you type.

peterix:
Could be a reset button in the same place.

Issue #3638, opened 2021-03-05
OpenSSL download links in BUILD.md are dead


Sebastian-byte:
System Information
-----------------------------
MultiMC version: Latest Develop (UTC: 21:25 PM 05/03/2021)

Operating System: Windows

Summary of the issue or suggestion:
----------------------------------------------
Windows OpenSSL download link it's broken [they now use GitHub](https://www.indyproject.org/2020/06/16/openssl-binaries-moved-to-github/). [There's a updated download link](https://github.com/IndySockets/OpenSSL-Binaries/raw/master/Archive/openssl-1.0.2g-i386-win32.zip), also many links should be updated to HTTP**S**.

What should happen:
------------------------------

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. Try to download OpenSSL from Windows build guide.

Suspected cause:
---------------------------
They now use GitHub for storing OpenSSL downloads.

Logs/Screenshots:
----------------------------
[//]: # (Please refer to https://github.com/MultiMC/MultiMC5/wiki/Log-Upload for instructions on how to attach your logs.)
![image](https://user-images.githubusercontent.com/62727185/110175319-8f407300-7dcf-11eb-9b5d-1dd30caafb12.png)

Additional Info:
---------------------------

PR #3648, opened 2021-03-09
Update link for OpenSSL for windows

theTisch21:
The old link that downloaded OpenSSL now returns a 404 error: https://indy.fulgan.com/SSL/Archive/openssl-1.0.2g-i386-win32.zip
I found a new link to this archive: https://github.com/IndySockets/OpenSSL-Binaries/tree/master/Archive
And I found this direct link to version 1.0.2g : https://raw.githubusercontent.com/IndySockets/OpenSSL-Binaries/master/Archive/openssl-1.0.2g-i386-win32.zip , which is the version needed. I have updated the BUILD.md file to use this new direct link


theTisch21:
Looks like #3638 beat me to this

Issue #3633, opened 2021-02-28
Twitch/Curse Modpack search not working properly


andry08:
System Information
-----------------------------
MultiMC version: 0.6.12-develop-1521

Operating System: Windows 10 20H2

Summary of the issue or suggestion:
----------------------------------------------
Searching for "Better Minecraft Modpack" in the twitch section leads to the wrong modpack being shown. Just for reference, this should be the correct one: https://www.curseforge.com/minecraft/modpacks/better-minecraft-modpack.
Note that this issue seems to be related to MultiMC, because searching with the api manually reports the correct modpack: 
https://addons-ecs.forgesvc.net/api/v2/addon/search?categoryId=0&gameId=432&index=0&pageSize=25&searchFilter=Better%20Minecraft&sectionId=4471&sort=0

What should happen:
------------------------------
The correct modpack should be listed

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. Open MultiMC and click  "Add Instance"

2. Go to the twitch section

3. Search for "Better Minecraft Modpack"

Suspected cause:
---------------------------
I suspect the issue to be something related to the url construction [here](https://github.com/MultiMC/MultiMC5/blob/4c62776044f3a7ad9efa3150ebec6150070502ed/application/pages/modplatform/twitch/TwitchModel.cpp#L162)

Logs/Screenshots:
----------------------------
[MultiMC.log](https://paste.ee/p/IkFnP#5GKkKXfn8vwfpneZHqGQkrsccZLIXUsB) doesn't seem to contain any useful info, but I'm including it just in case
[//]: # (Please refer to https://github.com/MultiMC/MultiMC5/wiki/Log-Upload for instructions on how to attach your logs.)

kb-1000:
From the code:
```cpp
"https://addons-ecs.forgesvc.net/api/v2/addon/search?"
        "categoryId=0&"
        "gameId=432&"
        //"gameVersion=1.12.2&"
        "index=%1&"
        "pageSize=25&"
        "searchFilter=%2&"
        "sectionId=4471&"
        "sort=0"
```
`%1` is the start index (initially 0), `%2` is the search term.
MultiMC does not apply any sorting on top.
Note that the URL you provided searches for "Better Minecraft", not "Better Minecraft Modpack".

andry08:
You're right, idk why, I meant to write "Better Minecraft".
Still, the problem is that searching for "Better Minecraft" on MultiMC yields different results from the api.
[...]

PR #3686, opened 2021-03-24
GH-3633 assume latest MC version, if CurseForge pack supports multiple

phit:
right now these are unused anyway, and it stops a bunch of packs from showing up in the search

fixes GH-3633

### DIFF
--- a/application/pages/modplatform/flame/FlameModel.cpp
+++ b/application/pages/modplatform/flame/FlameModel.cpp
@@ -263,11 +263,12 @@ void Flame::ListModel::searchRequestFinished()
             }
             pack.latestFile.addonId = pack.addonId;
             pack.latestFile.fileId = id;
-            // FIXME: what to do when there's more than one, or there's no version?
             auto versionArray = file.value("gameVersion").toArray();
-            if(versionArray.size() != 1) {
+            if(versionArray.size() < 1) {
                 continue;
             }
+
+            // pick the latest version supported
             pack.latestFile.mcVersion = versionArray[0].toString();
             pack.latestFile.version = file.value("displayName").toString();
             pack.latestFile.downloadUrl = file.value("downloadUrl").toString();


jamierocks (commented):
This seems a reasonable approach to this.

I am curious whether picking the first version in the array is always the latest Minecraft version though - or if the order is something not random?

phit:
Probably, but its their api, for all we know they could rename that key at any point in time too. I think really doesnt matter at all currently, that version isn't used anywhere in MultiMC not sure why we are even processing it.

jamierocks:
Huh, the way I figured it was we were using `pack.latestFile.mcVersion` for creating the Minecraft component later one.

phit:
nope, thats taken from the actual file thats downloaded only the download url is passed to the install task

https://github.com/MultiMC/MultiMC5/blob/0946c7c138d73a6e11835dc95da764ff1c0db5ad/application/pages/modplatform/twitch/TwitchPage.cpp#L104

application/pages/modplatform/twitch/TwitchModel.cpp
```diff
@@ -263,11 +263,12 @@ void Twitch::ListModel::searchRequestFinished()
             }
             pack.latestFile.addonId = pack.addonId;
             pack.latestFile.fileId = id;
-            // FIXME: what to do when there's more than one, or there's no version?
             auto versionArray = file.value("gameVersion").toArray();
-            if(versionArray.size() != 1) {
+            if(versionArray.size() > 0) {
```
peterix, line 267: This change is super weird. Are you sure this is correct?

This seems like exactly the opposite of what it should be.

peterix (changes requested):
1. I merged something that conflicts with this - namely a rename of everythign Twitch related
2. The logic of this change seems ... odd.

application/pages/modplatform/twitch/TwitchModel.cpp
phit: I messed up the direction when moving this from my Linux VM to my host PC to commit it, will fix and rebase.

PR #3715, opened 2021-04-05
NOISSUE Fedora build, ATL exceptions, and language tweaks

jamierocks:
- NOISSUE Fix build on Fedora 34
- NOISSUE Handle JSON exceptions in ATLauncher support


api/logic/translations/TranslationsModel.cpp
```diff
@@ -36,6 +36,9 @@ struct Language
         if(key == "pt") {
             locale = QLocale("pt_PT");
         }
+        else if (key == "en") {
```
jamierocks, line 39: It's probably worthwhile to have a shared method for this.

Issue #3536, opened 2021-01-17
Suggestion: make the loader mods list a treeview


necauqua:
So you can disable a mod and all of it's hard depencencies (addons or fixes or whatever), and just see them in a more organized fashion.

In the new Forge (1.13+) you can parse the depencencies from the `mods.toml` file and prior to that you'd have to actually read the bytecode for the @Mod annotation data and parse its `requires` field, so this might not be feasible for you to do sadly.

But this feature would be really cool so I am suggesting that for you to consider.

PR #3735, opened 2021-04-16
NOISSUE Parse META-INF/mods.toml for Forge 1.14+ mod metadata

kumquat-ir:
Enigmatica 6, before and after:
Screenshot_2021-04-16_13-42-03

${file.jarVersion} is what is stored in mods.toml, and seems to be replaced on runtime by Forge itself

This will add a new dependency, [toml11](https://github.com/ToruNiina/toml11), not sure if this should be in a submodule or raw code
Will also allow further parsing of mods.toml, such as for #3536

(this is my first time trying to contribute to a project, please let me know if i did anything wrong!)


Janrupf (approved):
Looks good to me. Not exactly sure what the policy on submodules is, it seems to be rather `whatever, everything goes`.

jamierocks:
> ${file.jarVersion} is what is stored in mods.toml, and seems to be replaced on runtime by Forge itself

Yep, `file.jarVersion` is the only supported subtitution at the moment though. The below is my handing in another project.

```go
			// Minecraft Forge supports substitutions in mods.toml files,
			// with data populated from the Jar's MANIFEST.
			// Substitutions are in the form ${file.KEY}.
			for _, mod := range forgeMods {
				if mod.Version == "${file.jarVersion}" {
					manifestVersion := manifest.Main["Implementation-Version"]
					if manifestVersion == "" {
						// This matches Minecraft Forge's behaviour
						manifestVersion = "NONE"
					}

					mod.Version = manifestVersion
				}
			}
```

kumquat-ir:
Added a somewhat ugly solution for that, but works just fine
Thanks for showing me where that information is actually located

peterix:
I prefer raw code. It's MIT licensed, so it can be a static library too... Except it's a header only c++ mess that will make build times even longer than they are now...

How about https://github.com/cktan/tomlc99 instead?

kumquat-ir:
That can be done.  Looks like the author of that already provides a wrapper for that library in c++ (https://github.com/cktan/tomlcpp), but that unfortunately seems to be in c++17.

kumquat-ir:
More verbose now (guess that's just what you get with C libraries), but definitely compiles faster, and works just as well

Issue #3287, opened 2020-08-11
Is there a way to set Minecraft parameters (not jvm)


20PercentRendered:
System Information
-----------------------------
MultiMC version: 0.6.12-custom

Operating System: Arch Linux 

Summary of the issue or suggestion: 
----------------------------------------------
A place to put Minecraft (not java) arguments, like server, port, proxy etc

What should happen:
------------------------------
A field to put minecraft arguments, the ones that go in the Main(String[] args) 

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1.  

2.

3.

Suspected cause:
---------------------------

Logs/Screenshots:
----------------------------
[//]: # (Please refer to https://github.com/MultiMC/MultiMC5/wiki/Log-Upload for instructions on how to attach your logs.)

Additional Info:
---------------------------
This is a suggestion

phit:
usecase?

20PercentRendered:
Joining servers on game launch, minecraft proxy, whatever args the main method supports.

PR #3804, opened 2021-05-22
Add ability to select a server to join in the instance settings

Janrupf:
This PR allows to set a server in the instance settings. Additionally it performs DNS lookup using the pattern `_minecraft._tcp.<server address>`, since Minecraft skips this step when using `--server` and `--port`.

I confirmed this to be working from Minecraft 1.0 up to the current version. Older version might work as well, given that they accept the required arguments using the applet wrapper.

Ref:
image

### DIFF
--- a/api/logic/BaseInstance.h
+++ b/api/logic/BaseInstance.h
@@ -147,7 +147,8 @@ public:
     virtual shared_qobject_ptr<Task> createUpdateTask(Net::Mode mode) = 0;
 
     /// returns a valid launcher (task container)
-    virtual shared_qobject_ptr<LaunchTask> createLaunchTask(AuthSessionPtr account) = 0;
+    virtual shared_qobject_ptr<LaunchTask> createLaunchTask(
+            AuthSessionPtr account, MinecraftServerTargetPtr serverToJoin) = 0;
 
     /// returns the current launch task (if any)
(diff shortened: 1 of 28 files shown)


Janrupf:
This is partially related to #3287, though only implements a part of it.

kb-1000:
Shouldn't that be _tcp and not _tpc?

Janrupf:
Oh yeah, it is... thats like the 10th time that I typo that today. Its correct in the code though :P

Janrupf:
New changes:

- Address and port have been merged and address parsing behaves now like within minecraft itself:
image

- Servers can now be joined (given the instance is not running already) directly from the servers page:
image

peterix:
This looks pretty good.

Janrupf:
Decided to add another feature: MultiMC now accepts `--server` (in combination with `--launch`) on the command line to launch an instance with a specified server. You can now make desktop shortcuts to start your favorite version with specified servers

AverageComet250:
Once this is done, can this be duplicated for worlds?

I'm guessing no since I'm pretty sure there isn't a Minecraft Param for that

Janrupf:
> I'm guessing no since I'm pretty sure there isn't a Minecraft Param for that
Yeah, sadly this will not be possible without mods

peterix:
Let's merge this and see what happens.

PR #3805, opened 2021-05-23
NOISSUE Flagged mods system

jamierocks:
This pull request introduces a post-installation dialog for getting user consent to have certain mods ('flagged mods') be enabled. At present this intends to cover mods for Tracking and Advertising purposes, and is limited to the modpacks.ch support - but should be rolled out to other platforms with time.

This pull request will disable any flagged mods that have not been explicitly allowed by the user, using the standard MultiMC disable mechanism - note that any disabled mods can always be enabled at a later date (through Instance Settings -> Loader Mods).

image

The above is an example dialog from FTB Cotton 1.0.0 where FTB Auxilium has been explictly enabled.

### DIFF
--- a/api/logic/modplatform/modpacksch/FTBPackInstallTask.cpp
+++ b/api/logic/modplatform/modpacksch/FTBPackInstallTask.cpp
@@ -9,8 +9,9 @@
 
 namespace ModpacksCH {
 
-PackInstallTask::PackInstallTask(Modpack pack, QString version)
+PackInstallTask::PackInstallTask(UserInteractionSupport *support, Modpack pack, QString version)
 {
+    m_support = support;
     m_pack = pack;
     m_version_name = version;
 }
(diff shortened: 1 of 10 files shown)


jamierocks:
I haven't gotten around to it yet - but I will experiment with flagging by mod id rather than file name.

The data could also be driven by remote meta (and could even be expanded by third-party lists, like uBlock) - which would allow mods to be introduced/removed/etc and be effected to more users quicker.

ImperatorStorm:
Seems like a really good idea

mdcfe:
Regardless of your stance on tracking, some packs may rely on mods that include "undesirable" behaviours to function as designed by the modpack authors. Furthermore, disabling certain mods at random may prevent players joining multiplayer servers that are supposedly running the "same" modpack.

Auto-disabling certain "undesirable" mods is a step too far for a general purpose launcher like MultiMC, as it inevitably *will* lead to users disabling mods without understanding the impact it may have on modpacks. MultiMC intentionally breaking modpacks by default through a dark pattern will make it a far less desirable launcher for any modpack users. 

This dialog should simply provide an easier means to *disable* these "undesirable" mods; it is far more sensible leave the boxes checked by default, and warn users of the ramifications of both keeping certain mods enabled *and* of disabling them in modpacks.

api/logic/modplatform/modpacksch/FTBPackInstallTask.cpp
```diff
@@ -174,6 +175,59 @@ void PackInstallTask::install()
...
+                flaggedMods.append(ModpacksCH::FlaggedMod {
+                    info.absoluteFilePath(),
+                    "FTB Auxilium",
+                    tr("Sends system information, crash reports, and when you join/quit singleplayer or multiplayer sessions to FTB's servers."),
+                    tr("Tracking")
+                });
+            }
+
+            // MCMC
+            if (info.fileName().startsWith("mcmc") && info.fileName().endsWith(".jar")) {
```
kashike, line 195: This is very vague.

JoelTroch: @jamierocks said the "filter by name" will be replaced by a "filter by mod ID". The "filter list" might not be part of MultiMC but externalized like "meta".

application/pages/modplatform/ftb/FtbFlaggedModsDialog.ui
```diff
@@ -0,0 +1,107 @@
...
+     </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QLabel" name="label_2">
+     <property name="text">
+      <string>These mods are disabled by default, you may override this below individually for each mod. These mods can later be enabled through the 'Loader Mods' settings page.</string>
```
mdcfe, line 30:
```suggestion
      <string>You may enable or disable any of these mods below individually. Note that disabling mods may alter or break modpack behaviour, and can interfere with online play. These mods can later be enabled or disabled through the 'Loader Mods' settings page.</string>
```

application/pages/modplatform/ftb/FtbFlaggedModsListModel.cpp
```diff
@@ -0,0 +1,95 @@
+#include "FtbFlaggedModsListModel.h"
+
+FlaggedModsListModel::FlaggedModsListModel(QWidget *parent, QVector<ModpacksCH::FlaggedMod> flaggedMods)
+    : QAbstractListModel(parent), m_flaggedMods(flaggedMods)
+{
+    // fill in defaults
+    for (const auto& mod : flaggedMods) {
+        m_modState[mod.fileName] = false;
```
mdcfe, line 8:
```suggestion
        m_modState[mod.fileName] = true;
```

mdcfe (changes requested):

api/logic/modplatform/modpacksch/FTBPackInstallTask.h
```diff
@@ -8,12 +8,39 @@
 
 namespace ModpacksCH {
 
+struct MULTIMC_LOGIC_EXPORT FlaggedMod {
```
Proximyst, line 11: Perhaps there should be a field with information as to what it tracks and how it may endanger the user?

Proximyst (changes requested):
While privacy is an important subject to me, potentially destroying modpacks for average users is an absolute no-go. That being said, there's the opening of an "uncheck everything" button as found in many check forms in relation to GDPR; perhaps this would be viable?

api/logic/modplatform/modpacksch/FTBPackInstallTask.h
Leo40Git: There's a `flagReason` field for that, though what it should contain is not exactly defined...

peterix:
I don't think I want this in this form at all.

Maybe it would be a good thing to have in a larger mod metadata system?

PR #3819, opened 2021-05-28
NOISSUE Support ATLauncher optional mods

jamierocks:
This is still very early in development, but this pull request serves to track it's progress.

screenshot of selection dialog


AverageComet250:
Rather than just for atlauncher, wouldn't this feature be good for all of them?

jamierocks:
This is a platform specific feature, no other platform I'm aware of support anyway.

AverageComet250:
Don't other platforms have optional mods too?

jamierocks:
modpacks.ch has it in metadata, but is unused - as far as I'm aware. There probably is some code that could be shared, but no other use for it yet.

In any case, no platform will have optional mod logic the same as ATLauncher does.

peterix:
Code looks good, I'm just going to merge this and see what happens.

That's what develop is for after all :)

PR #3825, opened 2021-05-31
Fix setup wizzard says no Java regardless

NewoIsTaken:
Hi everyone,

I noticed that the setup wizard always says "You do not have a working Java set up yet or it went missing"  even if the person does have Java installed. I have removed it.

Thanks!


kb-1000:
I do not see any fix for anything here?
Just an error message you made less detailed.

spannerman79:
I was going to say "is this something that one of my family did" cause whenever they have issues all I get out of them is "Oh it just crashed" - just like this new _updated_ error message would entail.

PR #3835, opened 2021-06-04
NOISSUE Added log info for online mode of instance

Zetabite:
This is the first iteration of implementation, for adding info about the online status of the instance, for debugging reasons, in the log.

### DIFF
--- a/application/LaunchController.cpp
+++ b/application/LaunchController.cpp
@@ -197,7 +197,7 @@ void LaunchController::launchInstance()
         return;
     }
 
-    m_launcher = m_instance->createLaunchTask(m_session);
+    m_launcher = m_instance->createLaunchTask(m_session, m_serverToJoin);
     if (!m_launcher)
     {
         emitFailed(tr("Couldn't instantiate a launcher."));
--- a/application/LaunchController.cpp
+++ b/application/LaunchController.cpp
@@ -15,6 +15,9 @@
 #include <minecraft/auth/YggdrasilTask.h>
 #include <launch/steps/TextPrint.h>
 #include <QStringList>
+#include <QHostInfo>
+#include <QList>
+#include <QHostAddress>
 
 LaunchController::LaunchController(QObject *parent) : Task(parent)
 {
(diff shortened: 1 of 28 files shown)


Zetabite:
Example log: https://paste.ee/p/pgHKE

kb-1000:
I'd also suggest adding the resolved IP address of the Mojang servers to the log, to make identifying alt service user easier

Zetabite:
That screams for an entire section for auth/online mode info

phit:
Technic has this https://github.com/TechnicPack/LauncherV3/blob/1a84e19be4ef962bc67dfae456e87e4b54dd3e98/src/main/java/net/technicpack/launcher/LauncherMain.java#L255
Which has shown to be very helpful to avoid helping people using illegitimate accounts and various other network problems.

Zetabite:
example log https://paste.ee/p/O9OTI

Zetabite:
https://paste.ee/p/9DRbw

application/LaunchController.cpp
```diff
@@ -215,7 +218,41 @@ void LaunchController::launchInstance()
...
+    } else {
+        online_mode = "offline";
+    }
+    QString auth_server_status;
+    if(m_session->auth_server_online) {
+        auth_server_status = "online";
+    } else {
+        auth_server_status = "offline";
+    }
+    m_launcher->prependStep(new TextPrint(m_launcher.get(), "Launchend instance in " + online_mode + " mode\nAuth server is " + auth_server_status + "\n", MessageLevel::MultiMC));
```
jamierocks, line 253: *launched

application/LaunchController.cpp
```diff
@@ -215,7 +218,41 @@ void LaunchController::launchInstance()
...
+    QStringList servers = {"authserver.mojang.com", "session.minecraft.net", "textures.minecraft.net", "api.mojang.com"};
+    QString resolved_servers = "";
+    QHostInfo host_info;
 
+    for(QString server : servers) {
+        host_info = QHostInfo::fromName(server);
+        resolved_servers = resolved_servers + server + " resolves to:";
+        if(host_info.addresses().size() > 0) {
+            for(QHostAddress address : host_info.addresses()) {
+                resolved_servers = resolved_servers + "\n" + address.toString();
```
jamierocks, line 231: IMO it would be nicer if the resolved servers were comma-separated.

Zetabite: could you explain what you mean, i dont quite understand in what way seperated, dont want to implement only for it to be not the way you thought off :D

jamierocks: Sure - take the following example:

```
textures.minecraft.net resolves to:
13.224.195.6
13.224.195.63
13.224.195.13
13.224.195.42
```

would become something like

```
textures.minecraft.net resolves to:
    [13.224.195.6, 13.224.195.63, 13.224.195.13, 13.224.195.42]
```

Zetabite: I can do that, that was actually almost what i planned to do the first time

Zetabite: with tab?

jamierocks: I think it looks clearer with a tab personally.

Zetabite:
#3849 maybe also add a message if instance was request online but no server was found, suggesting that either there is no connection to the internet or the mojang servers are down?

application/LaunchController.cpp
```diff
@@ -215,7 +218,46 @@ void LaunchController::launchInstance()
...
+                resolved_servers = resolved_servers + address.toString();
+                if(!host_info.addresses().endsWith(address)) {
+                    resolved_servers = resolved_servers + ", ";
+                }
+            }
+        } else {
+            resolved_servers = resolved_servers + "N/A";
+        }
+        resolved_servers = resolved_servers + "]\n\n";
+    }
```
peterix, line 240: Well, if we are aiming for offline mode, querying these could lead to a very long timeouts here. basically, if we want offline, skip this.

peterix (changes requested):
Also, please squash commits and use the proper commit message prefixes.

PR #3843, opened 2021-06-08
Update Java detection

phit:


jamierocks:
Are there other distributions we ought to introduce detection for too?

* Red Hat's
* Corretto
* Microsoft's
* ojdkbuild
* any more?

RDKRACZ:
What about Azul Zulu and Bellsoft Liberica?

DragonEggBedrockBreaking:
graalvm?

phit:
by default use the Oracle "JavaSoft" Path and are already picked up through that:

- Corretto
- ojdkbuild



---
added:

- Microsoft
- Bellsoft Liberica
- Azul Zulu

---

Red Hat: needs an account to download unless I'm blind
graalvm: doesn't seem to have an installer for Windows

jamierocks:
> Red Hat: needs an account to download unless I'm blind

It does, yeah.

PR #3870, opened 2021-06-18
NOISSUE Specify support URLs at build time

jamierocks:
Support URLs (bug tracker, Discord guild, subreddit) are now specified
as cache variables in cmake, and the buttons are not shown if no value
is set for them.

This is an early-stage move towards debranding the MultiMC codebase,
and will (hopefully) alleviate support requests coming to us from
illicit forks.


peterix (approved):

peterix:
Added the URLs as variables to TC, merging :)

Issue #3731, opened 2021-04-15
Limit vertical height of modpack versions selection to something reasonable


dexman545:
### Role

Playing with modpacks from Curseforge

### Suggestion

Limit vertical height of modpack version selection to something reasonable - it has scroll support, no need for it to to fill the entire screen.

### Benefit

Looks nicer, easier to view options, and wouldn't cover task/menu bar on windows as the selection does not close when the mouse is no longer over it.

### This suggestion is unique

I have searched the issue tracker and did not find an issue describing my suggestion, especially not one that has been rejected.

![image](https://user-images.githubusercontent.com/4873274/114799705-408adf00-9d66-11eb-869e-5225372d2f67.png)

jamierocks:
Yeah, we figured this could be a problem with some packs unfortunately :(

If possible, restricting the veritical height is indeed a good idea!

phit:
this is partially fixed now for anything that respects https://doc.qt.io/qt-5/qcombobox.html#maxVisibleItems-prop
for everything else it may be an idea to set `ui->versionSelectionBox->setStyleSheet("combobox-popup: 0;");`

but then the scrollbar is invisible on Windows at least ![](https://cdn.discordapp.com/attachments/134411126849929216/855502301870293002/unknown.png)

PR #3872, opened 2021-06-18
GH-3731 Limit Curserforge version height

phit:


phit:
after

image

before

image

PR #3894, opened 2021-06-22
NOISSUE: Rename Minecraft Log to Instance Log

Zetabite:
Renaming it, to avoid possible confusion with the actual minecraft logs, and more accurately describe what its true being is.

### DIFF
--- a/application/pages/instance/LogPage.h
+++ b/application/pages/instance/LogPage.h
@@ -38,7 +38,7 @@ public:
     virtual ~LogPage();
     virtual QString displayName() const override
     {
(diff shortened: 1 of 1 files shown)


peterix:
It's the only Minecraft log available, as far as I'm concerned.

peterix:
It's the stuff the game produces and prints into stdout/stderr.

Zetabite:
I guess. Just tried to make it more clear to the user, on the other hand, if you cant understand that, maybe you shouldnt be using mmc, considering its a power tool

PR #3896, opened 2021-06-23
NOISSUE Use MiB suffix for Java memory options

jamierocks:


peterix:
The very existence of this naming is ... annoying. It's a reaction to people who purposefully abused language for their own gain.

They should have been sued into the ground for false advertising and misleading their customers.

Oh well.

PR #3951, opened 2021-07-21
Fix Windows icon scaling issues

Tkain:
I don't actually know the reason as to why this occurs, but on the latest version of Windows, the MultiMC application icon scales incorrectly, only ever rendering up to 32x32 despite having sizes for 48x48 and 256x256 included in the file.

To fix this, I recreated the icon file using the official logo SVG files in GIMP, and it seems to render correctly at all sizes. The new icon appears visually identical to the old one at all resolutions (aside from _actually_ rendering correctly, that is). The new file is also only about half as large as the old one storage-wise.

See the image below as a demonstration of the fix:

Icon fix


Tkain:
Oops. Reopening with the correct branch name.

PR #3956, opened 2021-07-22
NOISSUE Disable Install Forge button on 1.17 (and above)

jamierocks:
It appears that Minecraft Forge for Minecraft 1.17 will require setting
JVM arguments to operate, this is not currently possible with MultiMC's
meta.

Therefore, when Forge for 1.17 is released a solution will need to be
looked into.


BenjaminHoegh:
What is the status on this anyway?

jamierocks:
> What is the status on this anyway?

It was merged.

jotoho:
> What is the status on this anyway?

The hardcoded block was removed again, if the description of this commit is accurate: 90a62c429a0031ff64113e07a2ee2bf5ce69fbbc

Issue #3857, opened 2021-06-12
Can't install MultiMC without installing Java 8.


RDKRACZ:
### Operating System

Linux

### Description of bug

MultiMC requires installing openjdk-8 even though I have openjdk-16 installed (on my main PC MC my 1.16 modpack runs fine with Java 16 and I don't see a reason to force this Java version especially when somebody would be playing only with MC 1.17).

### Steps to reproduce

Download the official package for Ubuntu and try to install it without `openjdk-8-jre` `openjdk-8-jre-headless`.

### This issue is unique

- [X] I have searched the issue tracker and did not find an issue describing my bug.

jamierocks:
It doesn't appear to be a hard dependency of the package -> https://github.com/MultiMC/MultiMC5/blob/9fafe3ffe6fd970b5096c31db8687e4c62c0cfa6/application/package/ubuntu/multimc/DEBIAN/control#L9

RDKRACZ:
By hard dependency, you mean that the installation can be avoided by using the --no-install-recommends flag for apt?
What if someone uses a GUI package installer like Gdebi? Wouldn't it be better to instead use the "Suggest" option (seems more valid now, after MC switched to Java 16)?

spannerman79:
_If_ it can be assumed that everyone will _only_ play MC 1.17+ then yes.

But that cannot be assumed for obvious reasons.

AverageComet250:
Or, 8 and 16 can both be set as the same type of dependency. Whether these are set as hard or soft dependencies is up to you, but I would say to set them _both_ as hard dependencies.

RDKRACZ:
Java 16 is the only version that can run versions prior to 1.17 and 1.17+. Also, MultiMC on Windows and Linux Generic does not install any Java so why it does with the deb package?

Or maybe set it like that: 16 as recommended and 8 as suggested.

Zetabite:
No. 16 can run some older versions, but as soon as your get mods into it, you will need java 8. and java 8 is what is recommend to use pre 1.17 for almost everything.

RDKRACZ:
Maybe I was lucky but I was able to run ~60 mods on Java 16 with MC 1.16.5. You also need at least Java 11 to run the C2ME mod.

jamierocks:
The rpm package didn't specify any, so perhaps just remove the recommendation?

kb-1000:
I probably just forgot that

rehashedsalt:
[...]

PR #3965, opened 2021-07-23
GH-3857: Remove openjdk-8-jre recommendation

Sebastian-byte:


kb-1000:
I'd wait with this until downloading Mojang Java runtimes is supported

kryptochaos:
Plus, jre 8 is needed for <1.17

Sebastian-byte:
> Plus, jre 8 is needed for <1.17

Well RPM package doesn't include any JDK, please read the issue for more context.

kb-1000:
Well, the RPM package doesn't do that simply because I forgot adding it.
And I'm also not too sure how I'd even express Recommends in an RPM.
Using that as a reason to justify this change isn't going to work well.

peterix:
Java 8 is required for all versions below 1.17

At some point, we will have a way for the launcher to download its own java runtimes... and then this change will be warranted.

For now, it is not.

jamierocks:
Perhaps the recommendation could be expanded to include Java 16 packages too then?

bryiewes:
The ubuntu repos dont have java 16 yet.

PR #3971, opened 2021-07-26
fix: lin-system layout

sam-kirby:
Properly propagate LINUX_DATA_DIR

I use `multimc-git` from AUR. 20b9f2b42a3b58b6081af271774fbcc34025dccb broke this in two ways.

1) Resources such as the logo moved.
2) `lin-system` layout mode stopped using the XDG_DATA_DIR by default.

This commit addresses the second issue.

### DIFF
--- a/launcher/CMakeLists.txt
+++ b/launcher/CMakeLists.txt
@@ -882,7 +882,7 @@ qt5_wrap_ui(MULTIMC_UI ${MULTIMC_UIS})
 qt5_add_resources(MULTIMC_RESOURCES ${MULTIMC_QRCS})
 
 # Add executable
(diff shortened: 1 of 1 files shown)


sam-kirby:
I understand you are about to debrand this repository and moving these sources to the logic component may be part of that.

If so, then this PR should probably be handled more like an issue - with the way things are currently setup, LINUX_DATA_DIR is not properly propagated, resulting in [this code](https://github.com/MultiMC/MultiMC5/blob/develop/launcher/MultiMC.cpp#L259-L264) being omitted, and so the default behaviour will be to use the directory the application is in as the data directory. This does not work if you intend to continue allowing the application to be shipped as a package which can be installed in the standard `/usr/bin` directory where a non-root user does not have write access.

PR #4054, opened 2021-08-29
Add flat icon for custom-commands

StaticRocket:
I noticed that there was an icon missing for the "Custom Commands" menu entry for most of the icon themes. I noticed the flat theme was just Google Material fonts recolored so I wen ahead an updated that.

@peterix You may want to update your icon themes at some point as they are also missing this icon.


peterix:
Hmm. I haven't made most of those - I made very few icons and most of them are instance icons or part of the main icon theme.

PR #4076, opened 2021-09-06
NOISSUE Detection of multiple custom theme folders

phit:


phit:
do not merge yet, need to add fallback to a stock theme when the selected theme was deleted to 
https://github.com/MultiMC/MultiMC5/blob/d644fb2094f623e45bff237ede7d432121f72072/launcher/MultiMC.cpp#L1145

ryleu:
You can make this a draft PR so that it doesn’t show as open.

PR #4103, opened 2021-09-15
Fix underaged account that isn't linked to a family help link.

tobfos2611:
Old link was 404.


RDKRACZ:
The part of the URL after "4403181904525" is unnecessary - the website will load without it just fine.

tobfos2611:
Okay, I'll remove all of that.

tobfos2611:
Done.

PR #4193, opened 2021-10-23
Include modpack version in the default instance title

jamierocks:
I've inadvertantly left some other commits here, I don't think they're too contentious for merging though - so I'll leave them.


peterix:
Yeah, this is nice.

PR #4221, opened 2021-11-03
Debrand README.md

WesleyBryie:


peterix:
While the README and BUILD documents do need some changes, this is not it.

Issue #4155, opened 2021-10-07
Move "Change group" button position


Rafii2198:
### Role

Multi-instance player

### Suggestion

Move "Change group" button in rmb context to second section

### Benefit

More intuitive placement.

### This suggestion is unique

- [X] I have searched the issue tracker and did not find an issue describing my suggestion, especially not one that has been rejected.

### You may use the editor below to elaborate further.

Basically the **first** section is just the read only name of the modpack but **second** section is where organization stuff like renaming, and changing icon are placed. It would make sense if changing group would be located in same section as it is purely organization stuff.
But currently "Change group" button is located in **4th** section which is for managing the instance itself which I think should be moved to the **2nd** section.

Screenshot for cotext:
![obraz](https://user-images.githubusercontent.com/26695327/136453155-394c9dad-c57a-4af4-959b-8f3c5c5b11bd.png)

PR #4234, opened 2021-11-08
GH-4155 Reorder change group action

NewoIsTaken:
In issue #4155, it was suggested to move the "Change Group" option up with the rename/change icon options. This should accomplish that suggestion.


peterix:
Fixed manually.

PR #4255, opened 2021-11-18
NOISSUE Change patreon button to "MultiMC"

Samisafool:
This PR changes the "launcher" in the patreon button to be MultiMC, because the button links to the MultiMC patreon.
Any forks can just revert this change if they change it to a different patreon.


peterix:
The right answer here is IMO to make the patreon thing completely optional, disabled by default, and a disctinct branding item.

Issue #4044, opened 2021-08-26
Screenshot Copy to Clipboard


AxoCode:
### Role

Both for modded and vanilla

### Suggestion

Copy to Clipboard button below the `Upload` button, in the Screenshots Tab. If you press it, its just copy the picture to your clipboard

### Benefit

So that people dont always upload to imgur and post it, or drag and drop. Just copy it to clipboard and post it

### This suggestion is unique

- [X] I have searched the issue tracker and did not find an issue describing my suggestion, especially not one that has been rejected.

PR #4300, opened 2021-12-03
GH-4044 Implemented copy screenshots to the clipboard

Ghosty141:
- Added context-menu entry
- Ctrl+C keybind works as well
- If multiple screenshots are selected, only the first one gets copied

One "issue" I noticed, should the switch-case that checks for the pressed key be rewritten to if-else since the check for Ctrl+C cannot be done using the switch?

### DIFF
--- a/launcher/ui/pages/instance/ScreenshotsPage.cpp
+++ b/launcher/ui/pages/instance/ScreenshotsPage.cpp
@@ -252,7 +252,7 @@ bool ScreenshotsPage::eventFilter(QObject *obj, QEvent *evt)
     QKeyEvent *keyEvent = static_cast<QKeyEvent *>(evt);
 
     if (keyEvent->matches(QKeySequence::Copy)) {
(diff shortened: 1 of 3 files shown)


peterix:
I'd call this 'copy image data', and only allow it with single selection.

With both multiple selection and single selection, there should be a 'copy' that operates on the file level, not image data level.

phit:
related issue https://github.com/MultiMC/Launcher/issues/4044

Ghosty141:
> I'd call this 'copy image data', and only allow it with single selection.
> 
> With both multiple selection and single selection, there should be a 'copy' that operates on the file level, not image data level.

Implemented in a97d0a3 and e9c52ec . What I noticed, the widebard still contains the item, can this also be changed based on the selection?

peterix:
It should be possible?

Sorry to keep you waiting, but I just got distracted and forgot.

In any case, let's merge this and see what the dev build users think.

PR #4321, opened 2021-12-10
Update Java version to 7+

davispuh:
Currently it doesn't build with newer Java unless I make this change.

Note that even for 7 I do get warning but atleast it builds fine :)
```
-- Found Java: /usr/bin/java (found suitable version "17.0.1", minimum required is "7") found components: Development 
warning: [options] bootstrap class path not set in conjunction with -source 7
warning: [options] source value 7 is obsolete and will be removed in a future release
warning: [options] target value 7 is obsolete and will be removed in a future release
```

### DIFF
--- a/libraries/javacheck/CMakeLists.txt
+++ b/libraries/javacheck/CMakeLists.txt
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 3.1)
 project(launcher Java)
-find_package(Java 1.6 REQUIRED COMPONENTS Development)
+find_package(Java 7 REQUIRED COMPONENTS Development)
 
 include(UseJava)
 set(CMAKE_JAVA_JAR_ENTRY_POINT JavaCheck)
-set(CMAKE_JAVA_COMPILE_FLAGS -target 1.6 -source 1.6 -Xlint:deprecation -Xlint:unchecked)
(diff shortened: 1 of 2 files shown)


theofficialgman:
I was already told that MultiMC will not change this requirement https://github.com/MultiMC/Launcher/issues/3949

> Java 8 is needed as the launch stuff is set to run even with Java 6, newer compilers may not support this. You can change this yourself MultiMC will not change that requirement. https://github.com/MultiMC/MultiMC5/blob/60b686f01458fc1f31109a0be71d790b8f7ec22b/libraries/launcher/CMakeLists.txt#L7

davispuh:
That's just retarded and sounds like very poor excuse. If you really want to support that old Java (which is EOL long time ago and doesn't receive security fixes anymore) then could add CMake flag to allow build with outdated Java - that way both would work.

jamierocks:
You're not required to use Java 1.6 for compiling, that's just the compatibility level target.

Changing that would be a breaking change, and my understanding is won't happen during this major version of MultiMC. There is little need to make such change however, as Java 1.8 is still widely supported and packaged.

peterix:
Honestly, at this point... Maybe this is just fine?

Everyone should be using Java 8 with all the old MC releases anyway. It just works.

peterix:
Let's do this.

PR #4348, opened 2021-12-13
add official packages to README

binex-dsk:
This PR adds instructions to install the official Arch Linux and Gentoo Linux packages, as well as some instructions for other distributions and operating systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,25 @@ MultiMC 5
 
 MultiMC is a custom launcher for Minecraft that focuses on predictability, long term stability and simplicity.
 
+## Packages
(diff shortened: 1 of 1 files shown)


Zetabite:
multimc-git is not official

Zetabite:
so by default, this pr is already not even in talk to get accepted

binex-dsk:
multimc-git is "official" in the sense that it is the ONLY available, up-to-date package which directly builds and installs this system-wide, not per-user like multimc-bin.

The same thing goes for the ebuild.

binex-dsk:
@peterix Please review, your other maintainers are unaware of the status of the official packages.

peterix:
Begone.

PR #4364, opened 2021-12-17
Change from Oracle JDK to AdoptOpenJDK

rfl890:
The Oracle JDK 8 is now offered at a commercial license and requires an account to download, which is not really ideal for most developers. AdoptOpenJDK has no such license (and most developers use it anyways) and is completely free to use.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -102,7 +102,7 @@ Getting the project to build and run on Windows is easy if you use Qt's IDE, Qt 
     - Microsoft Visual C++ 2008 Redist is required for this, there's a link on the OpenSSL download page above next to the main download.
     - We use a custom build of OpenSSL that doesn't have this dependency. For normal development, the custom build is not necessary though.
 * [zlib 1.2+](http://gnuwin32.sourceforge.net/packages/zlib.htm) - the Setup is fine
-* [Java JDK 8](https://adoptium.net/releases.html?variant=openjdk8&jvmVariant=hotspot) - Use the MSI installer.
+* [Java JDK 8](https://adoptium.net/releases.html?variant=openjdk8) - Use the MSI installer.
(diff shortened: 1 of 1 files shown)


jotoho:
AdoptOpenJDK is now [Adoptium](https://adoptium.net/)

BUILD.md
```diff
@@ -102,7 +102,7 @@ Getting the project to build and run on Windows is easy if you use Qt's IDE, Qt
     - Microsoft Visual C++ 2008 Redist is required for this, there's a link on the OpenSSL download page above next to the main download.
     - We use a custom build of OpenSSL that doesn't have this dependency. For normal development, the custom build is not necessary though.
 * [zlib 1.2+](http://gnuwin32.sourceforge.net/packages/zlib.htm) - the Setup is fine
-* [Java JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)
+* [Java JDK 8](https://adoptium.net/releases.html?variant=openjdk8&jvmVariant=hotspot) - Use the MSI installer.
```
jotoho, line 105: Does the `jvmVariant` argument even do anything? I tried accessing the website with and without that parameter and, as far as I can tell, it didn't change anything. Even the hashes for checking the files were identical on the win x64 msi download.

I think Adoptium only distributes their fancy "Temurin" and no longer ships hotspot like AdoptOpenJDK did. No idea what (if any) changes Temurin has in comparison to hotspot.

The URL can probably be shortened to `https://adoptium.net/releases.html?variant=openjdk8` without consequence.

rfl890: It specifies whether it uses OpenJ9 or Java HotSpot. I did notice it defaults to HotSpot.

PR #4373, opened 2021-12-19
Fixed small oversight in README.md

ImperatorStorm:
Launcher is open-source, MultiMC is not.


peterix:
Fair. But I think it needs a bigger rewrite.

I've been thinking of actually writing a thorough explanation of what's going on and the effects of the decisions made about the project.

I guess this is fine though. Let's merge it, I can always change it more.

Issue #4299, opened 2021-12-02
Uploading screenshot to imgur crashes multimc without error in log


ImperatorStorm:
### Operating System

Linux

### Description of bug

Attempting to [upload a screenshot to imgur via the "Screenshots" menu](https://user-images.githubusercontent.com/30777770/144520590-0777e8a3-3f20-46f8-9115-9b7adc134753.png) causes multimc to crash without any obvious errors.

[MultiMC-0.log](https://github.com/MultiMC/Launcher/files/7645780/MultiMC-0.log)

Using dev branch on [`multimc-bin`](https://aur.archlinux.org/packages/multimc-bin) on the AUR.

### Steps to reproduce

instance settings -> "Screenshots" -> select screenshot -> "Upload" -> silently crashes

### This issue is unique

- [X] I have searched the issue tracker and did not find an issue describing my bug.

NewoIsTaken:
I have some more information [here](https://paste.ee/p/FygcO#5lvqXv2tgxc4ohE00mF9QKlASaLCU3aS).
Specifically:
```
terminate called after throwing an instance of 'QString'
/usr/bin/multimc: line 23: 31517 Aborted                 (core dumped) ./MultiMC "$@"
```

ImperatorStorm:
stacktrace from `gdb`:

```
(gdb) backtrace
#0  0x00007ffff5d2cd22 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff5d16862 in abort () from /usr/lib/libc.so.6
#2  0x00007ffff60b6802 in __gnu_cxx::__verbose_terminate_handler () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x00007ffff60c2c8a in __cxxabiv1::__terminate (handler=<optimized out>) at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:48
#4  0x00007ffff60c2cf7 in std::terminate () at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_terminate.cc:58
#5  0x00007ffff60c2f8e in __cxxabiv1::__cxa_throw (obj=<optimized out>, tinfo=0xc89790 <typeinfo for QString>, dest=0x521570 <QString::~QString()>)
    at /build/gcc/src/gcc/libstdc++-v3/libsupc++/eh_throw.cc:95
#6  0x000000000057aaba in NetJob::start() ()
[...]

PR #4394, opened 2021-12-29
GH-4299 Fix Screenshot upload

Janrupf:
Uploading screenshots has been broken since commit https://github.com/MultiMC/Launcher/commit/69213b1206e97f7d4db4270a4b3b0af41dc9e6fc due to NetJob's requiring a network manager pointer.

This PR addresses 2 issues:
- Don't hard crash MultiMC in case the pointer is missing
- Set the pointer for screenshot tasks

An additional method for setting the network has been added as task dialog boxes and sequential tasks call `start` directly on tasks with no chance of setting the network.

### DIFF
--- a/launcher/net/NetJob.cpp
+++ b/launcher/net/NetJob.cpp
@@ -98,6 +98,12 @@ void NetJob::partProgress(int index, qint64 bytesReceived, qint64 bytesTotal)
 
 void NetJob::executeTask()
 {
+    if(!m_network) {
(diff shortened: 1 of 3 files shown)


phit:
I assume this addresses https://github.com/MultiMC/Launcher/issues/4299

Janrupf:
Good catch, didn't see the issue, I'll update the commit messages and PR title.

PR #4404, opened 2022-01-03
Fix OSI trademark issues in README.

HelenaMKG:
References directly connecting the project to MultiMC, which is technically a separate ARR project, may be legally problematic, as implying MultiMC itself is open-source may violate the trademark guidelines of the OSI for the usage of the term "open-source".

_[2.1](https://opensource.org/trademark-guidelines)_
> the use of the term "Open Source" is used solely in reference to software distributed under OSI Approved Licenses.

This README PR changes or removes any wording or references from the README that make the implications that may violate the trademark, makes some small clarifications on the MultiMC trademark usage, and is basically a CYA for the case of trademark infringement report-trolling that may happen to the project. Additionally, it may correct some misconceptions about MultiMC.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,11 @@
 <p align="center">
-  <img src="https://avatars2.githubusercontent.com/u/5411890" alt="MultiMC logo"/>
+  <img src="https://avatars2.githubusercontent.com/u/5411890" alt="Launcher logo"/>
 </p>
 
-MultiMC 5
+Launcher
 =========
 
-MultiMC is a custom launcher for Minecraft that focuses on predictability, long term stability and simplicity.
+Launcher is a custom launcher for Minecraft that focuses on predictability, long term stability and simplicity.
 
 ## Development
 If you want to contribute, talk to us on [Discord](https://discord.gg/multimc) first.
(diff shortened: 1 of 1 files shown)


Zetabite:
Since this is the official MultiMC repo, owned by MultiMC, i dont see us needing to remove the references to MultiMC

peterix:
The issue is imaginary.

- The license does not give you right to the name of the project.
- A fork can very easily strip out the few remaining occurences of the original name.
- The README is only there to be read by developers, who should know better anyway.
- What @Zetabite said.
- This is surface level cosmetics and doesn't change the situation in any way.

peterix:
On the pedantic side:
> We keep Launcher open source because we think it's important to be able to see the source code for a project like this, and we do so using the Apache license.

Does that claim that `MultiMC` is open source?
No.

Launcher is open source. MultiMC is a build of Launcher. The name may only be used in approved contexts. I haven't written a nice document for it aside from those few paragraphs in the README.

But the gist of it is this:
The name may not be applied to custom third party builds of any kind.
The name may not be applied to the launcher repackaged in a way that breaks its core functionality and confuses users.

peterix:
Ended up doing this instead:
https://github.com/MultiMC/Launcher/commit/addf5f4e52cf869375e2803afa0fa14dbaf71b22

It's clear enough.

PR #4461, opened 2022-01-28
Update rpm spec to support OpenSuse and conform to Fedora guidelines

Jan200101:


launcher/package/rpm/MultiMC5.spec
```diff
@@ -1,14 +1,20 @@
 Name:           MultiMC5
 Version:        1.4
-Release:        3%{?dist}
+Release:        4%{?dist}
 Summary:        A local install wrapper for MultiMC
 
 License:        ASL 2.0
 URL:            https://multimc.org
-BuildArch:      x86_64
+ExclusiveArch:  %{ix86} x86_64
```
Jan200101, line 8: `%{ix86}` includes all of `i386 i486 i586 i686 pentium3 pentium4 athlon geode`
I think these should work with the 32 bit tarball, but correct me if I am wrong

launcher/package/rpm/MultiMC5.spec
```diff
@@ -23,22 +29,29 @@ mkdir -p %{buildroot}/opt/multimc
 install -m 0644 ../ubuntu/multimc/opt/multimc/icon.svg %{buildroot}/opt/multimc/icon.svg
 install -m 0755 ../ubuntu/multimc/opt/multimc/run.sh %{buildroot}/opt/multimc/run.sh
 mkdir -p %{buildroot}/%{_datadir}/applications
-install -m 0644 ../ubuntu/multimc/usr/share/applications/multimc.desktop %{buildroot}/%{_datadir}/applications/multimc.desktop
+desktop-file-install --dir=%{buildroot}%{_datadir}/applications ../ubuntu/multimc/usr/share/applications/multimc.desktop
+
 mkdir -p %{buildroot}/%{_datadir}/metainfo
-install -m 0644 ../ubuntu/multimc/usr/share/metainfo/multimc.metainfo.xml %{buildroot}/%{_datadir}/metainfo/multimc.metainfo.xml
+install -m 0644 ../ubuntu/multimc/usr/share/metainfo/multimc.metainfo.xml %{buildroot}/%{_metainfodir}/multimc.metainfo.xml
```
Jan200101, line 35: `%{_metainfodir}` is the recommended macro for `/usr/share/metainfo`    
https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/

launcher/package/rpm/MultiMC5.spec
```diff
@@ -23,22 +29,29 @@ mkdir -p %{buildroot}/opt/multimc
...
 mkdir -p %{buildroot}/%{_mandir}/man1
 install -m 0644 ../ubuntu/multimc/usr/share/man/man1/multimc.1 %{buildroot}/%{_mandir}/man1/multimc.1
 
+%check
+appstream-util validate-relax --nonet %{buildroot}%{_metainfodir}/multimc.metainfo.xml
```
Jan200101, line 40: metainfo files need to be validated    
https://docs.fedoraproject.org/en-US/packaging-guidelines/AppData/#_app_data_validate_usage

launcher/package/rpm/MultiMC5.spec
```diff
@@ -23,22 +29,29 @@ mkdir -p %{buildroot}/opt/multimc
...
 %dir /opt/multimc
 /opt/multimc/icon.svg
 /opt/multimc/run.sh
 %{_datadir}/applications/multimc.desktop
-%{_datadir}/metainfo/multimc.metainfo.xml
-%dir /usr/share/man/man1
-%{_mandir}/man1/multimc.1.gz
+%{_metainfodir}/multimc.metainfo.xml
+%dir %{_mandir}/man1
+%{_mandir}/man1/multimc.1*
```
Jan200101, line 49: It is recommended to use a wildcard at the end since rpm is in charge of compressing the man page and in the future this might change from gzip to something else    
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_manpages

launcher/package/rpm/MultiMC5.spec
```diff
@@ -1,14 +1,20 @@
...
 Summary:        A local install wrapper for MultiMC
 
 License:        ASL 2.0
 URL:            https://multimc.org
-BuildArch:      x86_64
+ExclusiveArch:  %{ix86} x86_64
+
+BuildRequires:  desktop-file-utils
+BuildRequires:  libappstream-glib
+Requires:       zenity %{?suse_version:lib}qt5-qtbase wget xrandr
```
Jan200101, line 12: conditional logic requires `libqt5-qtbase` on Suse systems and `qt5-qtbase` on any other system

launcher/package/rpm/MultiMC5.spec
```diff
@@ -23,22 +29,29 @@ mkdir -p %{buildroot}/opt/multimc
 install -m 0644 ../ubuntu/multimc/opt/multimc/icon.svg %{buildroot}/opt/multimc/icon.svg
 install -m 0755 ../ubuntu/multimc/opt/multimc/run.sh %{buildroot}/opt/multimc/run.sh
 mkdir -p %{buildroot}/%{_datadir}/applications
-install -m 0644 ../ubuntu/multimc/usr/share/applications/multimc.desktop %{buildroot}/%{_datadir}/applications/multimc.desktop
+desktop-file-install --dir=%{buildroot}%{_datadir}/applications ../ubuntu/multimc/usr/share/applications/multimc.desktop
```
Jan200101, line 32: usage of desktop-file-install is demanded to ensure .desktop safety and spec compliance    
https://docs.fedoraproject.org/en-US/packaging-guidelines/#_desktop_file_install_usage

launcher/package/rpm/MultiMC5.spec
```diff
@@ -1,14 +1,20 @@
 Name:           MultiMC5
 Version:        1.4
-Release:        3%{?dist}
+Release:        4%{?dist}
 Summary:        A local install wrapper for MultiMC
 
 License:        ASL 2.0
 URL:            https://multimc.org
-BuildArch:      x86_64
```
Jan200101, line 8: ExclusiveArch is the better alternative when trying to limit what arch a package is used on.

Since this only packages assets and a bash script it might be worthwhile to set this to noarch instead of removing it entirely

launcher/package/rpm/MultiMC5.spec
```diff
@@ -1,14 +1,20 @@
...
+ExclusiveArch:  %{ix86} x86_64
+
+BuildRequires:  desktop-file-utils
+BuildRequires:  libappstream-glib
+Requires:       zenity %{?suse_version:lib}qt5-qtbase wget xrandr
+Provides:       multimc = %{version}
+Provides:       MultiMC = %{version}
+Provides:       multimc5 = %{version}
```
Jan200101, line 15: need to explictly define the version of what we provide to ensure anything or anyone depending on a specific version will get the right version.
It is recommended to also use %{?_isa} but since this package is practically arch independent that doesn't matter    
https://docs.fedoraproject.org/en-US/packaging-guidelines/#renaming-or-replacing-existing-packages

Jan200101 (commented):
here is a list of the reasoning behind each changes and the guidelines if they apply

peterix (commented):
Seems sensible.

What I really want to do is eventually move to packaging with https://github.com/pacur/pacur or something similar.

You start with what looks like a PKGBUILD and add the distribution quirks as special cases.

But that's the future. Improvements to the existing things are definitely welcome :)

launcher/package/rpm/MultiMC5.spec
kb-1000: As far as I'm aware, Suse systems don't define that macro (at least I had problems getting a package that refers to it to build there)

PR #4496, opened 2022-02-06
NOISSUE Attempt to make exit codes more useful on Windows

Janrupf:
This PR attempts to add more information to exit codes.

When Windows terminates an application (or the applications is hung **and** killed through taskmanager) the exit code is set to an `NTSTATUS`. The Launcher now attempts to decode those exit codes into an `NTSTATUS` and a status message.

Known problems:
- some messages contain format specifies MultiMC can't fill in, thus the messages will be a bit messy (still readable)
- in order to map to `NTSTATUS` codes the `ntstatus.h` header is read at compile time and inspected, maybe there is a better way
- when the JVM catches a crash, the exit code doesn't signal anything useful (however, the JVM itself prints the status value, so this should be fine)

Overall this should aid in debugging when the JVM doesn't catch a crash. Leaving this as a draft for now so there is some space for discussions.

In order to test this code https://github.com/Janrupf/crash-it may be used to forcefully crash Minecraft with different reasons.


Janrupf:
Alright, I feel like this can be reviewed, didn't find any breaking flaws

afkvido (approved):

peterix (changes requested):
Can you remove the parsing of the header file and code generation?
I'd be much more comfortable with a static lookup table and instructions how to update it instead.

PR #4599, opened 2022-03-23
NOISSUE count swaptotal in total ram (linux)

minishd:
Factors in swaptotal on top of memtotal when getting total ram amount on linux. Sorry if there is any issues, this is my first contribution


peterix:
Hmm. Are you sure this is a good idea?
The game would run really badly if you system is swapping. Practically, you want to use less memory than available RAM. The upper limit is not usable because of all the other things taking up memory.

peterix:
I mean, I guess this is fine, but maybe this could be exposed in the UI in some way. Like having a checkbox 'allow swap' that expands the range.

yccb:
You're right, I'll try to make a UI option for it.

peterix:
Please do not close pull requests like that.

PR #4637, opened 2022-04-11
New download method 

AnasSae-ed:


peterix:
I'm pretty sure this is illegal, or promotes illegal actions.

PR #4664, opened 2022-05-02
NOISSUE Support installing CurseForge packs through modpacks.ch

jamierocks:
This makes the neccasery changes to the modpacks.ch pack installation code to supporting installing CurseForge packs thorugh the platform, but doesn't make any of the changes to the GUI to support doing this.


peterix:
Hmmm closed. Why? Is this not viable at all?

PR #4671, opened 2022-05-06
NOISSUE Add Install Legacy Fabric Button

### LINKED ISSUE #4408: An "Install Legacy Fabric" button.
### Role

I would like to play with Legacy Fabric on MultiMC.

### Suggestion

On applicable instances (<=1.13.2) allow the user to select the button, on newer versions, gray it out.

### Benefit

Because it would make playing Legacy Fabric easier. 

### This suggestion is unique

- [X] I have searched the issue tracker and did not find an issue describing my suggestion, especially not one that has been rejected.

### You may use the editor below to elaborate further.

Btw their website is here at https://bruh.legacyfabric.net.

I can see this being implemented in multiple ways.

Ungray the Install Fabric button and allow the user to select it, but download legacy fabric instead.

Make a new Install Legacy Fabric button that is grayed on 1.14+

Note: Legacy Fabric doesn't mean older versions of Fabric, it means that it's Fabric for <=1.13.2.

thecatcore:
To merge along side MultiMC/meta#9.
Closes #4408

### DIFF
--- a/launcher/minecraft/ComponentUpdateTask.cpp
+++ b/launcher/minecraft/ComponentUpdateTask.cpp
@@ -591,7 +591,7 @@ void ComponentUpdateTask::resolveDependencies(bool checkOnly)
                     {
                         component->m_version = "3.1.2";
                     }
-                    else if (add.uid == "net.fabricmc.intermediary" || add.uid == "org.quiltmc.hashed")
+                    else if (add.uid == "net.fabricmc.intermediary" || add.uid == "org.quiltmc.hashed" || add.uid == "net.legacyfabric.intermediary")
                     {
                         auto minecraft = std::find_if(components.begin(), components.end(), [](ComponentPtr & cmp){
                             return cmp->getID() == "net.minecraft";
--- a/launcher/ui/pages/instance/VersionPage.cpp
+++ b/launcher/ui/pages/instance/VersionPage.cpp
@@ -213,7 +213,7 @@ void VersionPage::updateVersionControls()
     auto minecraftVersion = Version(m_profile->getComponentVersion("net.minecraft"));
 
     bool supportsFabric = minecraftVersion >= Version("1.14");
-    ui->actionInstall_Fabric->setEnabled(controlsEnabled && supportsFabric);
+    ui->actionInstall_Fabric->setEnabled(controlsEnabled);
     ui->actionInstall_Quilt->setEnabled((controlsEnabled) && supportsFabric);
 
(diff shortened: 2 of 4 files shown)


kb-1000:
I disagree with how this is done, you're making it confuse Fabric with legacy Fabric.
They're not the same thing, and shouldn't be treated as such.
See https://github.com/MultiMC/Launcher/issues/4408#issuecomment-1006104947

arthurbambou:
Is it really required when basically it installs upstream FabricLoader + Legacy Fabric Intermediary? 🤔

PseudoDistant:
> I disagree with how this is done, you're making it confuse Fabric with Legacy Fabric. They're not the same thing, and shouldn't be treated as such. See [#4408 (comment)](https://github.com/MultiMC/Launcher/issues/4408#issuecomment-1006104947)

> They're not the same thing
Dude they use exactly the same loader, the only thing different is the intermediary file.

jamierocks:
The button reads 'Install Fabric', now I would take that to mean Fabric - not a fork there-of.

This should be implemented as Quilt has just been, not by masquerading a third-party project as the upstream one.

PseudoDistant:
@jamierocks you misunderstand.
It's not a fork, we literally use upstream Fabric Loader.
The only difference is the Intermediary being fed to it.

jamierocks:
I don't misunderstand, I'm well aware of what Legacy Fabric is and isn't - and what Fabric is and isn't.

The button doesn't read 'Install Fabric Loader', it reads 'Install Fabric' - it currently does as I would expect and installs the Fabric Loader, and requires Fabric's intermediaries. These are part of the package that Fabric's installer would install, and together really form part of what most would consider Fabric.

Installing intermediaries from a third-party isn't what I would consider installing Fabric, it would be installing **Legacy Fabric**. Just because the package happens to use Fabric Loader, doesn't make the package what most would understand to be Fabric.

arthurbambou:
Updating accordingly to submitted suggestions.

arthurbambou:
bump

arthurbambou:
oops fixed a typo...

JFronny:
Forgive me if I read this wrong, but wouldn't it make more sense to not disable the button but hide it on newer versions and hide fabric on older ones?
Since neither will ever be available on versions where the other exists and they work as de-facto "equivalents" for another, adding a disabled button seems to me like unneeded clutter and this approach also avoids the problems mentioned by jamierocks.

arthurbambou:
How would I do it?

ImperatorStorm:
Why was this closed?

PR #4718, opened 2022-05-24
NOISSUE Patreon button always links to MultiMC

Sebastian-byte:
Another possible solution would be to give an option in CMake build files for a Patreon link.

### DIFF
--- a/launcher/ui/MainWindow.cpp
+++ b/launcher/ui/MainWindow.cpp
@@ -403,8 +403,8 @@ public:
         actionPatreon = TranslatedAction(MainWindow);
         actionPatreon->setObjectName(QStringLiteral("actionPatreon"));
         actionPatreon->setIcon(APPLICATION->getThemedIcon("patreon"));
-        actionPatreon.setTextId(QT_TRANSLATE_NOOP("MainWindow", "Support %1"));
-        actionPatreon.setTooltipId(QT_TRANSLATE_NOOP("MainWindow", "Open the %1 Patreon page."));
+        actionPatreon.setTextId(QT_TRANSLATE_NOOP("MainWindow", "Support MultiMC"));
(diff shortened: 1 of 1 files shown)


peterix:
I did this work on making sure the project can be easily forked, so people would leave me alone...

jamierocks:
I think the better solution here would be to add the CMake option, and keep this as is - holds better to the idea of debranding.

peterix:
No, I am not interested in splintering the branding into multiple locations.

peterix:
If anything, this should be part of the branding, and conditional on the presence of the setting.

peterix:
That is, these should be in `notsecrets` / `secrets`:
image

peterix:
The other option is to actually undo the debranding again... Because the one project that could have done the right thing, didn't. And this whole idea of debranding kind of failed.

jamierocks:
> Because the one project that could have done the right thing, didn't.

I thi

peterix:
Well, parametrizing all of these is consistent at least.

peterix:
But I don't see anything good coming from the debranding after the experience with what came after.

peterix:
On its own, pulling out the magical variables into a common place so they don't have to be kept track of all over the codebase is good.

jamierocks:
> But I don't see anything good coming from the debranding

I think in retrospect, it would have helped if there was an open-source branding variant of the project (like Chromium to Chrome), so that MultiMC was exclusive to multimc.org and 'Multium' could be used by packagers. Launcher is too generic and combined with the logo, screams 'replace me' - which was it's intent, I know.

> On its own, pulling out the magical constants into a common place so they don't have to be kept track of all over the codebase is good.

Agreed :)

peterix:
Anyway, that would be worth a PR

PR #4733, opened 2022-05-28
NOISSUE Enforce allowed domains in Modrinth packs

arthomnix:
Adds a check to make sure that download URLs in Modrinth packs are on an allowed domain as per the [Modrinth pack specification](https://docs.modrinth.com/docs/modpacks/format_definition/#downloads).

This will likely break a lot of current Modrinth packs as it does not include CurseForge CDN domains that were recently removed from the specification upon CurseForge's request. It may be advisable to delay merging until Modrinth has removed packs with CF download URLs from their website.

### DIFF
--- a/launcher/InstanceImportTask.cpp
+++ b/launcher/InstanceImportTask.cpp
@@ -314,7 +314,7 @@ void InstanceImportTask::processModrinth() {
                     throw JSONValidationError("Download URL for " + file.path + " is not a correctly formatted URL");
                 }
                 QString downloadDomain = file.download.host();
-                if (std::find(std::begin(Modrinth::allowedDownloadDomains), std::end(Modrinth::allowedDownloadDomains), downloadDomain) == std::end(Modrinth::allowedDownloadDomains))
+                if (!Modrinth::allowedDownloadDomains.contains(downloadDomain))
                 {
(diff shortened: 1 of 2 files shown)


launcher/modplatform/modrinth/ModrinthPackManifest.h
```diff
@@ -21,6 +21,14 @@
 #include <QUrl>
 
 namespace Modrinth {
+
+const QString allowedDownloadDomains[] = {
```
peterix, line 25: Use `QStringList` instead.

launcher/InstanceImportTask.cpp
```diff
@@ -313,6 +313,11 @@ void InstanceImportTask::processModrinth() {
                 {
                     throw JSONValidationError("Download URL for " + file.path + " is not a correctly formatted URL");
                 }
+                QString downloadDomain = file.download.host();
+                if (std::find(std::begin(Modrinth::allowedDownloadDomains), std::end(Modrinth::allowedDownloadDomains), downloadDomain) == std::end(Modrinth::allowedDownloadDomains))
```
peterix, line 317: Use https://doc.qt.io/qt-5/qstringlist.html#contains instead of this magic.

peterix (changes requested):

arthomnix:
Closing as Modrinth have clarified that the domain restrictions are specific to their platform rather than the pack format as a whole

PR #4808, opened 2022-07-03
NOISSUE (Re-)implement the ability to create instance shortcuts

arthomnix:
Adds the ability to create a desktop shortcut to launch an instance, using MMC's command line arguments.
This was apparently a MMC4 feature that has been missing in MMC5 for 10 years.

~~This is currently WIP and only works on Linux, and is missing a few features such as setting an icon for the shortcut automatically.~~ Now works cross-platform and is ready for review

### DIFF
--- a/launcher/CMakeLists.txt
+++ b/launcher/CMakeLists.txt
@@ -778,7 +778,8 @@ SET(LAUNCHER_SOURCES
     ui/dialogs/VersionSelectDialog.h
     ui/dialogs/SkinUploadDialog.cpp
     ui/dialogs/SkinUploadDialog.h
-
+    ui/dialogs/CreateShortcutDialog.cpp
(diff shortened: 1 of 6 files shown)


arthomnix:
Now has support for Windows and Mac, going to mark this ready for review

arthomnix:
This should actually be ready now

kb-1000:
Do you really have support for mac? I can't see that in the code

arthomnix:
It's the same as for Linux but with the option to create a shell script force-enabled (because macs don't have proper shortcuts)

arthomnix:
although i don't actually have a mac so if anybody does testing would be appreciated

peterix:
I'll just merge this and let's see how/if it breaks.

PR #4822, opened 2022-07-09
NOISSUE log file extension of mods to clear ambiguity

xcfrg:
- While troubleshooting an odd issue with someone on discord that ended up being browsers downloading mods with incorrect file extensions, I noticed that MultiMC did not log file extensions for the mods folder
- Logging this information will make it much easier to pick up on these kinds of mistakes
- How it looks:

Y0708_233344


Janrupf:
I really like this, lgtm

Issue #4812, opened 2022-07-03
[Feature] Copy suggested instance name


mk-pmb:
### Role

Playing a Vanilla plus modpack.

### Suggestion

MultiMC should help me make minor adjustments to my to-be-created instance's name.

### Benefit

I'd have to type less.

### This suggestion is unique

- [X] I have searched the issue tracker and did not find an issue describing my suggestion, especially not one that has been rejected.

### You may use the editor below to elaborate further.

When I select the modpack for my new instance, MultiMC suggests a name for the instance, and it's almost the one I want. However, I can neither modify it directly, nor copy it and paste it into a text editor to modify there, because the suggested name is not "real" text.

Could you insert the suggestion as actual text? To continue to allow people who want to start typing their own name right away, you could auto-select the text on focus.

Alternatively, could you make it so that attempts to copy the empty name field will instead copy the suggested name?

Alternatively, could you do that when in the name field, one of the key combinations is pressed that in the active OS we usually expect to mean "copy to clipboard"? (on Windows and Ubuntu, that would be Ctrl+C and Ctrl+Insert.)

mk-pmb:
Thank you all for adding this!

PR #4828, opened 2022-07-11
GH-4812 Prefill instance name to allow making adjustments

arthomnix:
Fills the instance name in instead of just setting a placeholder. This allows adjustments to be made to the suggested name without typing the whole thing out.
The text is selected by default so that typing will overwrite the text, but users who want to adjust the default name instead of typing their own can deselect the text.
The placeholder name is still set so it is still visible if the user deletes the text.
Also sets the focus to the instance name textbox by default, whereas previously it was on the group name - this is required so the text gets overwritten on typing but also makes more sense generally.
Closes issue #4812.

### DIFF
--- a/launcher/ui/dialogs/NewInstanceDialog.cpp
+++ b/launcher/ui/dialogs/NewInstanceDialog.cpp
@@ -154,10 +154,10 @@ void NewInstanceDialog::setSuggestedPack(const QString& name, InstanceTask* task
     creationTask.reset(task);
 
     defaultInstName = name;
+    ui->instNameTextBox->setPlaceholderText(name);
 
     if (!instNameChanged)
     {
-        ui->instNameTextBox->setPlaceholderText(name);
         ui->instNameTextBox->setText(name);
     }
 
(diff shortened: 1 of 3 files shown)


launcher/ui/dialogs/NewInstanceDialog.cpp
```diff
@@ -151,6 +151,9 @@ void NewInstanceDialog::setSuggestedPack(const QString& name, InstanceTask* task
 {
     creationTask.reset(task);
     ui->instNameTextBox->setPlaceholderText(name);
+    ui->instNameTextBox->setText(name);
+    ui->instNameTextBox->selectAll();
+    ui->instNameTextBox->setFocus();
```
peterix, line 156: Can you do it without the setFocus?

peterix (changes requested):

launcher/ui/dialogs/NewInstanceDialog.cpp
peterix: Detecting already made changes and not resetting the name could also be a thing, along with a reset button that sets the text to the current suggested name and clears that flag.

arthomnix: Done all of this, let me know if there is a better way to detect if the user has clicked inside the text box

PR #4845, opened 2022-07-20
NOISSUE Shortcut creation bugfixes

arthomnix:
This means that the script will run when clicked, instead of being opened in a text editor


arthomnix:
Added a couple more shortcut creation bugfixes

arthomnix:
added an edge case fix (MMC being installed in a path containing literal double quotes)
(only relevant on Linux and Mac since Windows does not allow quotes in paths)

PR #4912, opened 2022-09-12
Sandbox MultiMC launcher

cumpootuhr:
The main motivation for this was the (not-so-)recent log4j vulnerability. I'd like a way to restrict what someone can do if they manage to successfully infect my system through a Minecraft/Java vulnerability, especially if I intend to play online on older Minecraft versions that will likely never see a log4j fix. While I'm aware that MultiMC already has a patched version of log4j, I think it makes sense to additionally sandbox the entire launcher. This has the additional benefit of mitigating potential future exploits that we don't know about yet

Note that this adds an additional dependency to the launcher: Bubblewrap. This is unavoidable if we want to use Bubblewrap to sandbox the game

### DIFF
--- a/launcher/Launcher.in
+++ b/launcher/Launcher.in
@@ -14,7 +14,7 @@ if [[ $EUID -eq 0 ]]; then
 fi
 
 
-LAUNCHER_NAME=MultiMC
+LAUNCHER_NAME=@Launcher_Name@
 LAUNCHER_DIR="$(dirname "$(readlink -f "$0")")"
 echo "Launcher Dir: ${LAUNCHER_DIR}"
 
--- a/launcher/Launcher.in
+++ b/launcher/Launcher.in
@@ -14,7 +14,7 @@ if [[ $EUID -eq 0 ]]; then
 fi
 
 
-LAUNCHER_NAME=@Launcher_Name@
+LAUNCHER_NAME=MultiMC
 LAUNCHER_DIR="$(dirname "$(readlink -f "$0")")"
 echo "Launcher Dir: ${LAUNCHER_DIR}"
 
@@ -29,24 +29,26 @@ export QT_FONTPATH="${LAUNCHER_DIR}/fonts"
 # Detect missing dependencies...
 DEPS_LIST=`ldd "${LAUNCHER_DIR}"/plugins/*/*.so 2>/dev/null | grep "not found" | sort -u | awk -vORS=", " '{ print $1 }'`
 if [ "x$DEPS_LIST" = "x" ]; then
-    # We have all our dependencies. Run the launcher.
-    echo "No missing dependencies found."
+    if [ "$1" != "--dry-run" ]; then
+        # We have all our dependencies. Run the launcher.
+        echo "No missing dependencies found."
 
-    # Just to be sure...
-    chmod +x "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}"
+        # Just to be sure...
+        chmod +x "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}"
 
-    # Run the launcher
-    exec -a "${LAUNCHER_DIR}/${LAUNCHER_NAME}" "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}" -d "${LAUNCHER_DIR}" "$@"
+        # Run the launcher
+        exec -a "${LAUNCHER_DIR}/${LAUNCHER_NAME}" "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}" -d "${LAUNCHER_DIR}" "$@"
 
-    # Run the launcher in valgrind
-    # valgrind --log-file="valgrind.log" --leak-check=full --track-origins=yes "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}" -d "${LAUNCHER_DIR}" "$@"
+        # Run the launcher in valgrind
+        # valgrind --log-file="valgrind.log" --leak-check=full --track-origins=yes "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}" -d "${LAUNCHER_DIR}" "$@"
 
-    # Run the launcher with callgrind, delay instrumentation
-    # valgrind --log-file="valgrind.log" --tool=callgrind --instr-atstart=no "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}" -d "${LAUNCHER_DIR}" "$@"
-    # use callgrind_control -i on/off to profile actions
+        # Run the launcher with callgrind, delay instrumentation
+        # valgrind --log-file="valgrind.log" --tool=callgrind --instr-atstart=no "${LAUNCHER_DIR}/bin/${LAUNCHER_NAME}" -d "${LAUNCHER_DIR}" "$@"
+        # use callgrind_control -i on/off to profile actions
 
-    # Exit with launcher's exit code.
-    # exit $?
+        # Exit with launcher's exit code.
+        # exit $?
+    fi
 else
     # apt
     if which apt-file &>/dev/null; then
--- a/launcher/package/rpm/MultiMC5.spec
+++ b/launcher/package/rpm/MultiMC5.spec
@@ -9,7 +9,7 @@ ExclusiveArch:  %{ix86} x86_64
 
 BuildRequires:  desktop-file-utils
 BuildRequires:  libappstream-glib
-Requires:       zenity %{?suse_version:lib}qt5-qtbase wget xrandr
+Requires:       zenity %{?suse_version:lib}qt5-qtbase wget xrandr bubblewrap
 Provides:       multimc = %{version}
 Provides:       MultiMC = %{version}
 Provides:       multimc5 = %{version}
--- a/launcher/package/ubuntu/multimc/DEBIAN/control
+++ b/launcher/package/ubuntu/multimc/DEBIAN/control
@@ -5,7 +5,7 @@ Maintainer: Petr Mrázek <peterix@gmail.com>
 Section: games
 Priority: optional
 Installed-Size: 75
-Depends: zenity, desktop-file-utils, libqt5widgets5, libqt5gui5, libqt5network5, libqt5core5a, libqt5xml5, libqt5concurrent5, wget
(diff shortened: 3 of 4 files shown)


Zetabite:
Is this even supported cross-platform, like, does it natively work on Windows?

theofficialgman:
> While I'm aware that MultiMC already has a patched version of log4j, I think it makes sense to additionally sandbox the entire launcher.

actually no, it makes zero sense to sandbox the entire launcher given the "exploits" you wish to prevent. the launcher and the jvm which it launches running minecraft are separate processes. If you are concerned about exploits running in minecraft, you would wish to sandbox the jvm.

> This has the additional benefit of mitigating potential future exploits that we don't know about yet

I don't see how that acomplishes really anything, you give access to the whole user binary space in this PR (PATH) which you of course have to do in order for the jvm to actually function and run the game. So all you accomplished was not giving access to the users HOME directories effectivly. You haven't prevented any exploits in bubblewrap or the underlying kernel or libraries that would allow escaping the sandbox.

cumpootuhr:
> actually no, it makes zero sense to sandbox the entire launcher given the "exploits" you wish to prevent. the launcher and the jvm which it launches running minecraft are separate processes. If you are concerned about exploits running in minecraft, you would wish to sandbox the jvm.

By sandboxing the launcher, any child processes are created with the same restrictions. This results in the JVM having the same restrictions as the launcher itself.

> I don't see how that acomplishes really anything, you give access to the whole user binary space in this PR (PATH) which you of course have to do in order for the jvm to actually function and run the game. So all you accomplished was not giving access to the users HOME directories effectivly

This is true with the current way it's setup. I included documentation in the source file explaining how I'd like to implement stronger protections, but at the current config still presents some significant benefits:
- The user's home directory is restricted completely
  - This prevents an attacker from getting read access to user files without a sandbox escape
  - This prevents trivial privilege escalation on administrator accounts such as fakesudo attacks
- All capabilities held by the current user are dropped
  - This reduces what an attacker can do given malware running inside the sandbox
- The entire filesystem is mounted read-only
  - This prevents someone from modifying world-writable or user-writable data outside of the user's home directory as a means to escalate privilege or move laterally

As for the threat of sandbox escapes and kernel exploitation, this *is* a significant issue, although it's worth noting that Bubblewrap refuses to grant SUID permissions to any child processes spawned from inside the sandbox, leaving kernel exploits and IPC attacks as the primary attack surface. This is already a huge improvement against the threat model of most Minecraft players, which is people using off-the-shelf exploits to attack the lowest common denominator of users. Of course it won't protect you much against targeted attacks unless your attacker is a script kiddie, but it's always possible to tighten restrictions.

I eventually plan to restrict permissions more granularly, as explained in the documentation accompanying my changes. But even by simply denying home directory access we cut off the most significant threat to the largest portion of players, and force attackers trying to target MultiMC users to pair their initial exploit with a subsequent sandbox escape, making MultiMC players less appealing targets.

ImperatorStorm:
You should also modify [`launcher/package/ubuntu/multimc/DEBIAN/control`](https://github.com/MultiMC/Launcher/blob/develop/launcher/package/ubuntu/multimc/DEBIAN/control) and [`launcher/package/rpm/MultiMC5.spec`](https://github.com/MultiMC/Launcher/blob/develop/launcher/package/rpm/MultiMC5.spec) to include `bubblewrap` as a dependency.

theofficialgman:
> By sandboxing the launcher, any child processes are created with the same restrictions. This results in the JVM having the same restrictions as the launcher itself.

yeah that is my point, that is bad.
to add to my statements earlier:

it is actually worse to put the launcher and the jvm in the same sandbox and here is why. MultiMC stores account tokens in plain text in the accounts.json file within the launcher's storage directory (~/.local/share/multimc). unsandboxed or sandboxed the way you have it, if there was a vulnerability that could remotely execute code on the system, you would have access to this file (mojang and microsoft credentials included). And I am sure that this isn't the only file we wouldn't want the jvm to have access to.

This basically closes this PR as it is not useful to sandbox the launcher and jvm together. if you want usefull sandboxing, you will need to implement it on the jvm only.

cumpootuhr:
The point of the sandboxing isn't to protect login tokens, for 1 simple reason: The client itself has access to a user's login session anyway, since the launcher needs to pass relevant login data to the client for the client to be able to play online. Session stealing is the go-to method of temporarily stealing accounts anyway, and it can performed without any access outside the JVM itself by someone who manages to get code execution within the JVM.

The point of the sandboxing is to protect a user's system from an exploitable vulnerability affecting the rest of their system. With this PR, infections are contained within Minecraft, instead of having unrestricted access to the system. I fail to see the merit in any argument that this isn't unequivocally good for player security, when it clearly restricts the movement of malware on a victim machine. If you want to complain about my PR not sandboxing the JVM as well as you'd like, you should submit your own PR making this change.

peterix:
1. I do not want this stuff as a dependency or requirement. For example, my system does not and **will not** contain the stuff this requires to function.
2. It affects the launcher and it integrating properly with the system
3. It is not granular enough, and it is done at the wrong layer. Sandboxing instances is the right way to do things.
4. This only works on Linux, a minority OS. It does nothing whatsoever to protect other 95% of the users, most of which are not mega nerds like the linux community, and are more susceptible to getting malware.

So, the general idea is OK, but this has about 0% chance of getting merged.

peterix:
On top of that, the amount of assumptions this makes about the system and how it works is unmanintainable:
```
    bwrap --die-with-parent --cap-drop all --new-session --unshare-all --share-net \
        --tmpfs / --tmpfs /tmp --tmpfs /home --tmpfs /root --dev-bind /dev /dev --proc /proc \
        --ro-bind /usr/bin /usr/bin --bind /bin /bin --ro-bind-try /etc/alternatives /etc/alternatives \
        --ro-bind /lib /lib --ro-bind-try /lib64 /lib64 --ro-bind-try /lib32 /lib32 \
        --ro-bind /usr/lib /usr/lib \
        $path_fwd --bind "$INSTDIR" "$INSTDIR" \
        --ro-bind-try /run/user/"$UID"/pulse /run/user/"$UID"/pulse \
        --ro-bind-try /run/user/"$UID"/pipewire-0 /run/user/"$UID"/pipewire-0 \
        --ro-bind-try /tmp/.X11-unix/X0 /tmp/.X11-unix/X0 \
        --ro-bind-try /run/user/"$UID"/wayland-1 /run/user/"$UID"/wayland-1 \
        --ro-bind /etc/fonts /etc/fonts \
        --ro-bind /usr/share/icons /usr/share/icons \
        --ro-bind /usr/share/mime /usr/share/mime \
        --ro-bind /usr/share/fontconfig /usr/share/fontconfig \
        --ro-bind /usr/share/fonts /usr/share/fonts \
        --ro-bind /etc/resolv.conf /etc/resolv.conf \
        --ro-bind /etc/ssl /etc/ssl \
        --ro-bind /etc/ca-certificates /etc/ca-certificates \
        --ro-bind /usr/share /usr/share \
    ./MultiMC "$@"
```

peterix:
Do note that right now, MultiMC is meant to work on every linux distro from 2014 onwards. Was this available in 2014?

peterix:
A more realistic approach from my POV would be to deeply integrate sandboxing into the launcher.

In general, I'd want to also change how the JVM sees the instance folder(s) and isolate the instance from some undesirable things the mods do. Like trying to mess with the launcher.

There's also a case for virtualizing the filesystem access entirely, and for example mapping a shared (all instances) and local (the sandboxed instance) resource folders into something that appears as one to the game, and writes go to the local one.

peterix:
So, in its current form, I am simply going to close this.

cumpootuhr:
You said the sandboxing is implemented in the wrong layer, although I explained why that isn't considered a major issue. Bur as far as that, I'll concede that it could implement slightly more fine-grained permissions than sandboxing the entire launcher (proposal: nested sandboxing? But the launcher in a weaker sandbox, and the JVM in a stronger one. It wouldn't be hard to implement that)

As far as the amount of system assumptions being unmaintainable, you can easily just merge the original version of this PR which simply `--ro-bind`s the entire rootfs, but seeing as that in itself received complaints, I feel like I'm put in a catch 22. If I strengthen the isolation, I get complaints because it makes too many assumptions. But if I weaken the isolation, it's not considered effective isolation. How exactly is anything supposed to improve if we impose this kind of lose-lose system?

Also, I fail to see how it matters that Bubblewrap was implemented after 2014. It's not 2014 anymore, Modern Minecraft versions recommend using Java 17, which didn't exist in 2014 either. If you haven't updated your system since 2014, you also can't access a large number of websites because they dropped support for your browser years ago.

As far as sandboxing being optional, that's easy. I've already pushed a commit implementing this suggestion. It has 3 sandboxing modes:

1. Full sandboxing. For systems with Bubblewrap installed that can launch a sandbox without issue
2. Reduced sandboxing. For systems with Bubblewrap that can't launch a full sandbox due to the "unmaintainable" set of system assumptions, this `--ro-bind`s `/`, drops capabilities, unshares namespaces, unshares home directories, and calls it a sandbox
3. No sandboxing. On systems without Bubblewrap, it just disables sandboxing entirely

It tells the user what form of sandboxing it's using, then waits 1 second before starting MC. This gives the user a chance to know if their instance is being sandboxed without piping the output to a file.

cumpootuhr:
I don't see why you felt the need to close this PR, since most of the changes you requested are so trivial to fix. And if someone implements JVM sandboxing later, that can be integrated without even removing the launcher sandbox (as I mentioned, nested sandboxing is possible. It just requires us to weaken the parent sandbox a bit so that is has the permissions necessary to create a child sandbox)

cumpootuhr:
Oh, I forgot to discuss Bubblewrap being Linux-specific. This is sort of inherent if we want any form of sandboxing, since I'm unaware of any sandbox that's cross-platform across Linux, Windows, and Mac (even just supporting Linux and BSD is hard, since Linux, FreeBSD, and OpenBSD all require separate sandboxing mechanisms under-the-hood). You'd need separate code paths for spawning sandboxes across major OSs. But that's ok, since this change was implemented in a shell script anyway, which itself only exists on a minority of systems. I might also cite the number of issues requesting Flatpak distribution, despite Flatpak also only existing on Linux systems. Sandboxing is clearly a feature people want, even if it only works on some systems. Just because some people can't benefit from a feature doesn't mean it shouldn't be implemented.

cumpootuhr:
> It affects the launcher and it integrating properly with the system

I specifically went out of my way to add binds to allow the launcher to integrate with the system. If it doesn't integrate properly, that's a problem with the specific rules I chose and can be worked out through testing

PR #4929, opened 2022-10-03
NOISSUE Quote Instance ID when creating a shortcut

AnthonyPorthouse:
When a shortcut is created if the instance folder contains spaces, the instance will fail to launch from the shortcut as it does not correctly handle the spaces in the `-l` argument.

To fix this issue we should correctly quote the Instance ID value.


Janrupf:
Basically duplicate of #4845

AnthonyPorthouse:
Missed that since I only searched issues, Happy to close in favour of #4845.

PR #4964, opened 2022-11-02
NOISSUE Add setting to display playtime in hours only

arthomnix:
This feature has been requested a lot on Discord as other game launchers such as Steam display time in hours only.


1991mateus1991-cmyk:
ERIC

PR #5032, opened 2023-01-06
When re-logging in when the repair account expires, it still prompts that the account has expired

graytoowolf:


Zetabite:
So what is it that your PR does? Neither title, nor a comment nor branch name are stating that

graytoowolf:
> So what is it that your PR does? Neither title, nor a comment nor branch name are stating that

When re-logging in when the repair account expires, it still prompts that the account has expired

arthomnix:
> So what is it that your PR does? Neither title, nor a comment nor branch name are stating that

Looks like it fixes a bug in one of my PRs (#4998) which meant that after re-logging into your account it would still give the expired error message
As the person who wrote the code that this PR is changing, this should be merged

PR #5052, opened 2023-02-04
GH-4699 Modrinth pack exporter

### LINKED ISSUE #4699: Modrinth Output Support
### Role

I make Forge and Fabric modpacks

### Suggestion

Export instances in the Modrinth modpack format

### Benefit

so that way MultiMC will have a way to upload instances to Modrinth

### This suggestion is unique

- [x] I have searched the issue tracker and did not find an issue describing my suggestion, especially not one that has been rejected.

### You may use the editor below to elaborate further.

_No response_

arthomnix:
Adds the ability to export instances to the Modrinth .mrpack modpack format.
~~Currently very WIP and missing features.~~ Should be ready now!

Fixes #4699


kb-1000:
You might want to look for a datapacks folder. While this isn't supported by vanilla, neither are shaderpacks, and it's not uncommon to use a mod adding support in modpacks. Perhaps more common than using Iris or Optifine in modpacks even.

arthomnix:
@kb-1000 Is there any standard as to what the global datapacks folder is called? Every mod I can find seems to use a different folder

arthomnix:
I've ended up just letting the user specify the global datapacks folder

arthomnix:
This should actually be ready now, I'm just slightly worried about not having stable anymore

peterix:
Well, let's merge this and fix if it misbehaves?

PR #5063, opened 2023-02-08
NOISSUE Set Modrinth export dialog window title 

arthomnix:


launcher/net/Download.cpp
```diff
@@ -122,13 +122,6 @@ void Download::downloadError(QNetworkReply::NetworkError error)
         qCritical() << "Aborted " << m_url.toString();
         m_status = Job_Aborted;
     }
-    else if(error == QNetworkReply::ContentNotFoundError && (m_options & Option::AllowNotFound))
-    {
-        // The Modrinth API returns a 404 when a hash was not found when performing reverse hash lookup, we don't want to treat this as a failure
-        qDebug() << "Received 404 from " << m_url.toString() << ", continuing...";
-        m_status = Job_Finished;
-        return;
-    }
```
peterix, line 131: Are you sure this is unused?

peterix: (I'm not, just thinking, you added this for a reason, right?)

arthomnix: I added it when I was using the Modrinth API endpoint to lookup a single hash, the endpoint to lookup multiple hashes which it uses now doesn't return 404s

PR #5108, opened 2023-03-26
Skin Allocating doesn't work

TheFarlandsExplorer15:
In old MultiMC, there was an working skin in the stable versions, but now we have an skin allocating issue with the .png files in Minecraft, FIX is allocating the skin in the memory of the worlds (even though we might get an slightly larger world files in older versions, especially inf-20100531).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,33 @@
-# MultiMC 0.6.16
+# MultiMC 0.7.0
+
+This is a nice big rollup release with all kinds of things that accumulated since the last release.
+
+With this, we are also abandoning the concept of stable releases. From now on, MultiMC will always use the `develop` channel.
+
+Some selected changes:
+
+- Remove concept of switching update channels
+- Improve path parsing for modrinth packs
+- GH-4978: Add button to log back in when the account has expired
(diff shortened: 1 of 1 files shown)


changelog.md
```diff
@@ -1,4 +1,33 @@
-# MultiMC 0.6.16
+# MultiMC 0.7.0
+
+This is a nice big rollup release with all kinds of things that accumulated since the last release.
+
+With this, we are also abandoning the concept of stable releases. From now on, MultiMC will always use the `develop` channel.
```
TheFarlandsExplorer15, line 5: Why do you always need to have an "develop" channel when a old stable release I have (now develop).

changelog.md
```diff
@@ -1,4 +1,33 @@
-# MultiMC 0.6.16
+# MultiMC 0.7.0
```
TheFarlandsExplorer15, line 1: What about 0.6.17?!

TheFarlandsExplorer15 (commented):
Why do we need these things?!

phit:
I have no idea what you want to achieve with this pull request, if you have general questions about the project please come ask on our discord.

PR #5124, opened 2023-04-08
NOISSUE Fix directly joining servers on 23w14a and above

arthomnix:
Minecraft 23w14a removed the `--server` and `--port` command line options and replaced them with a new `--quickPlayMultiplayer` option. This makes MultiMC pass the new option on instances with 23w14a and higher, fixing the `-s` command line flag, and shortcuts with the "join server on launch" option enabled, when launching these versions.


peterix:
Ideally, we would put this as a tag/property on the release instead of having heuristics in the client, but let's go with this for a while longer.

PR #5125, opened 2023-04-08
NOISSUE Add support for launching worlds directly via Quick Play

arthomnix:
Adds support for launching directly into a singleplayer world using 23w14a's new Quick Play feature.
This can be done in 3 ways:
* Via new "Join" and "Join Offline" buttons on the Edit Instance > Worlds screen (enabled only on 23w14a and higher instances)
* Via an instance setting, similar to the one for servers (again, only enabled on 23w14a and higher)
* Via a new `--world`/`-w` CLI option (this is mutually exclusive with `--server`; no validation is performed on the Minecraft version but the help text specifies that it only works on 23w14a and higher)

Stuff still to do:
- [x] Hide the Join and Join Offline buttons on Edit Instance > Worlds if the version is too old, instead of just disabling them
- [x] Add a "join world on launch" option to the shortcut creation dialog

### DIFF
--- a/launcher/ui/pages/instance/WorldListPage.cpp
+++ b/launcher/ui/pages/instance/WorldListPage.cpp
@@ -320,12 +320,6 @@ void WorldListPage::worldChanged(const QModelIndex &current, const QModelIndex &
     bool enable = index.isValid();
     ui->actionJoin->setVisible(enableJoinActions);
     ui->actionJoinOffline->setVisible(enableJoinActions);
-    // The above lines will only hide the actions from the right click menu, to hide them from the toolbar we need to do this.
-    // If any more actions are added above the Join/Join Offline actions, these indices will need to be updated!
-    ui->toolBar->actions().at(2)->setVisible(enableJoinActions);
-    ui->toolBar->actions().at(3)->setVisible(enableJoinActions);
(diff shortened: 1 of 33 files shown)


launcher/ui/pages/instance/WorldListPage.cpp
```diff
@@ -311,8 +313,21 @@ void WorldListPage::mceditState(LoggedProcess::State state)
...
     QModelIndex index = getSelectedWorld();
     bool enable = index.isValid();
+    ui->actionJoin->setVisible(enableJoinActions);
+    ui->actionJoinOffline->setVisible(enableJoinActions);
+    // The above lines will only hide the actions from the right click menu, to hide them from the toolbar we need to do this.
+    // If any more actions are added above the Join/Join Offline actions, these indices will need to be updated!
+    ui->toolBar->actions().at(2)->setVisible(enableJoinActions);
+    ui->toolBar->actions().at(3)->setVisible(enableJoinActions);
+    // Also disable the separator so there aren't 2 separators in a row
+    ui->toolBar->actions().at(4)->setVisible(enableJoinActions);
```
arthomnix: This works, but I feel like there is a better way than hardcoding the indices?

PR #5162, opened 2023-05-22
NOISSUE Fix shortcut dialog bugs

arthomnix:
Fixes bugs in the create shortcut dialog where the OK button and server address text box were always disabled on instances that don't support Quick Play.

### DIFF
--- a/launcher/ui/dialogs/CreateShortcutDialog.ui
+++ b/launcher/ui/dialogs/CreateShortcutDialog.ui
@@ -1,9 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<!--
- Copyright 2022-2023 arthomnix
- This source is subject to the Microsoft Public License (MS-PL).
- Please see the COPYING.md file for more information.
--->
 <ui version="4.0">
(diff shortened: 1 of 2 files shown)


launcher/ui/dialogs/CreateShortcutDialog.ui
```diff
@@ -454,6 +454,22 @@
...
+    <hint type="sourcelabel">
+     <x>140</x>
+     <y>59</y>
+    </hint>
+    <hint type="destinationlabel">
+     <x>397</x>
+     <y>164</y>
+    </hint>
+   </hints>
+  </connection>
```
peterix: Please do not add more of these, connect the signals directly in code

peterix (changes requested):

launcher/ui/dialogs/CreateShortcutDialog.ui
peterix: And ideally using the new style syntax, like here:
image

arthomnix:
(accidentally committed changes intended for #5163 to the wrong branch)

PR #5201, opened 2023-06-26
NOISSUE Update Debian/Ubuntu package

arthomnix:
Makes the Debian package download the develop version of MultiMC since stable is no longer updated. Also makes the package recommend Java 17 as well as 8 since newer game versions require it.

The Arch pkgbuild repo should also be updated when this is merged.


ImperatorStorm:
Should also recommend `openjdk-21-jre`, as its required by [`>=24w14a`](https://www.minecraft.net/en-us/article/minecraft-snapshot-24w14a)

PR #5253, opened 2023-09-04
Fixed Wrong Proxy Settings Message

0xIO32:
image
It says "Minecraft does not accept proxy settings.", but thats wrong:
image
(Decompiled 1.20.1 using jd-gui)

### DIFF
--- a/launcher/ui/pages/global/ProxyPage.ui
+++ b/launcher/ui/pages/global/ProxyPage.ui
@@ -39,7 +39,7 @@
        <item>
         <widget class="QLabel" name="proxyPlainTextWarningLabel_2">
          <property name="text">
-          <string>This only applies to the launcher. Minecraft does not accept proxy settings.</string>
+          <string>This only applies to the launcher.</string>
          </property>
          <property name="alignment">
           <set>Qt::AlignCenter</set>


Zetabite:
iirc, it just doesn't do anything with em.

0xIO32:
Let me verify that

0xIO32:
It does use it for authentication:
image
image

arthomnix:
That isn't MultiMC in the screenshot

0xIO32:
It's a fork of it and its the same on MultiMC:
image

peterix:
I don't think this is particularly interesting to change. It is not used to play the game. It also doesn't apply to all game versions.

peterix:
Like yeah, we can change the string... But I'd rather do something more substantial. Like encode which versions actually take these args in metadata.

0xIO32:
> I don't think this is particularly interesting to change. It is not used to play the game. It also doesn't apply to all game versions.

Yes, but it technically spreads misinformation.
Maybe i will create a pr (or an issue, i never coded with c++...) to add it to metadata and create an entry in the installation config for passing proxy arguments 🤔.
But for now the string annoys me.

Issue #4356, opened 2021-12-15
[V0.6.14-develop-3019 on Win32][Windows 10 64] Launcher not detecting internet access after system sleep


Jahfry:
### Operating System

Windows

### Description of bug

Launcher should detect internet connection coming back up similar to how it detects when down. 

### Steps to reproduce

1) Run current Dev release of MultiMC on Windows 10 64

2) Launch an instance

3) [optional] Close instance from 2

4) Keep launcher running

5) "Sleep" the Windows machine (I've been doing overnight, unknown how long is needed, probably fairly instant)

6) "Wake" machine / login

7) Launch an instance

... at this point, step 7, even after 30 minutes, MultiMC still prompts for offline username (which it never does in step 2)

Not a big deal, just filing to get it off of my brain every time it happens :)

### This issue is unique

- [X] I have searched the issue tracker and did not find an issue describing my bug.

Khaeops:
I have this problem as of public version 0.6.14, always prompts me for my offline username when launching the game from sleep/hibernation. Have to close and restart MultiMC for it to connect again.

p-lucero:
At least for me, this appears to be a DNS issue - specifically, an inability to resolve `authserver.mojang.com`. I've grabbed [a snippet of my logs](https://gist.github.com/p-lucero/e81b806257753bbec66073e649f06c0d), and the absolute event timestamps (`2021-12-30 00:28:21-0700` and `2021-12-30 10:45:37-0700`) seem to line up with my last two wake-from-sleep occurrences.

Possible cause: on my machine, the network adapter takes a few seconds to come online after wake-from-sleep, which is likely what causes the name resolution to fail. However, this did only start happening with the most recent release (upgrading from `0.6.13` --> `0.6.14` stable releases on Win10) so there should be a corresponding code change that also caused this behavior.

(Side note: would it be acceptable to modify the MultiMC logger to output absolute timestamps on each log line, rather than timestamps relative to program start? I find the absolute timestamps much more useful for debugging :])

ejm554:
Yeah, I'm having this trouble too, but I have (incorrectly?) attributed it to when my Windows machine loses access to its internet connection sometime while the app was running, i.e., if the internal WiFi dropped its connection. I would expect that re-establishing a WiFi connection would allow me to use "Launch" without seeing the offline prompt. I, too, can workaround the problem by restarting MultiMC. I'm seeing this problem with Version 0.6.15-develop-3227 on win 32.

[...]

Issue #5166, opened 2023-05-28
Continues to ask for offline mode username after reconnecting to the internet


Lamby777:
### Operating System

Windows

### Description of bug

So, when you're offline, and you try to play, it'll automatically try to launch in Offline Mode, so it asks for your username, right? If you just realized then that you didn't have internet, and you back out of the menu and try again after connecting, it still asks for your offline mode name. This might seem kinda like an edge case, but it happened to me for something like the 5th time today, so I might as well report it.

Not sure if this is too similar to #4356, but that one's mostly focused on sleep rather than not having connection altogether, so yea.

### Steps to reproduce

1. Disconnect from the internet
2. Open MultiMC
3. Launch an instance
4. X out of the popup asking for your offline mode username
5. Connect to internet (don't close MultiMC)
6. Launch the instance again
(shows same popup)

### Suspected cause

Haven't looked at the code, but I'd assume it's checking for internet when you open up MultiMC, and then just remembers that you don't have internet as long as that process is open? Maybe check for internet at launch-time? idk

### This issue is unique

- [X] I have searched the issue tracker and did not find an issue describing my bug.

PR #5284, opened 2023-10-21
GH-5166 Adding user account state to be refreshed during Instance Launch

SithumDissanayake:
In the login function of LaunchController, added requestRefresh on the current m_accountToUse. This allows the accountState of accountToUse to be updated before running through what login is meant to depending on the accountState. This is to allow for the cases where a device might be back online after launching the Launcher. Putting this in the tryagain while loop also allows to check if a users account is expired during the login process.

Adds a force account refresh during the login process, updating the account state before the program performs an action depending on the account state. 
This allows for users to start an instance with no internet connection, realise and update their internet connection and start an instance again with no issue.
Putting this in the tryagain while loop also allows to check if a users account is expired during the login process.
This also resolves issue GH-4356.
Tested as much as I could on Ubuntu Linux 22.04.3 LTS. No noticeable lag or crashes noticed with multiple launches rapidly in a row and was able to confirm that the account status does change in this case.

### DIFF
--- a/launcher/LaunchController.cpp
+++ b/launcher/LaunchController.cpp
@@ -117,6 +117,11 @@ void LaunchController::login() {
 
     while (tryagain)
     {
+        // Force account refresh on the account used to launch the instance,
+        // updating the AccountState everytime between instance launch
+        auto accounts = APPLICATION->accounts();
+        accounts->requestRefresh(m_accountToUse->internalId());
+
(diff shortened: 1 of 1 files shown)


peterix:
It should really wait on the refresh, and only do the refresh if the account is soft-errored (offline, not online in general)

peterix:
I'll do it differently, it's on the current todo list.

PR #5315, opened 2023-12-31
Simplify Microsoft Login

kaduvert:
removes the need to enter the code when opening the login link by adding it as the `otc` query parameter to the url


phit:
there's instances where you don't login on the same device as multimc, please keep the code visible

PR #5323, opened 2024-01-10
GH-5232 Add Install NeoForge button

### LINKED ISSUE #5232: NeoForge Support Tracking Issue
This issue serves as a tracking issue for adding support for [NeoForge](https://neoforged.net), a fork of MinecraftForge which for the moment maintains compatibility but may diverge in the future. Given the current maintained compatibility and possible future divergence, plus the new state of the NeoForged project, this seems the perfect time to work out how compatibility will be handled and add support in the launcher.

arthomnix:
Add a button to install NeoForge.
Depends on https://github.com/MultiMC/ForgeWrapper/pull/1 and https://github.com/MultiMC/meta/pull/18. Closes #5232.

(Untested obviously since there is no NeoForge metadata yet.)

I would also like to fix the very broken code to detect whether Fabric/Quilt are supported (which currently doesn't disable the Install Fabric/Quilt buttons on pre-1.0 versions), and add NeoForge support detection, before merging.

### DIFF
--- a/launcher/minecraft/VersionFilterData.cpp
+++ b/launcher/minecraft/VersionFilterData.cpp
@@ -70,4 +70,7 @@ VersionFilterData::VersionFilterData()
     java16BeginsDate    = timeFromS3Time("2021-05-12T11:19:15+00:00");
     java17BeginsDate    = timeFromS3Time("2021-11-16T17:04:48+00:00");
     quickPlayBeginsDate = timeFromS3Time("2023-04-05T12:05:17+00:00");
+    liteLoaderEndsDate  = timeFromS3Time("2017-09-18T08:39:46+00:00");
+    fabricBeginsDate    = timeFromS3Time("2019-04-23T14:52:44+00:00");
+    neoForgeBeginsDate  = timeFromS3Time("2023-09-20T09:02:57+00:00");
 }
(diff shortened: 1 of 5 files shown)


arthomnix:
I've rewritten modloader support detection based on release date checks (like are currently used for detecting whether your Java version is too old), but it would probably be better to move these checks (and probably everything else in `VersionFilterData`) into meta at some point.

launcher/minecraft/VersionFilterData.cpp
```diff
@@ -70,4 +70,7 @@ VersionFilterData::VersionFilterData()
     java16BeginsDate    = timeFromS3Time("2021-05-12T11:19:15+00:00");
     java17BeginsDate    = timeFromS3Time("2021-11-16T17:04:48+00:00");
     quickPlayBeginsDate = timeFromS3Time("2023-04-05T12:05:17+00:00");
+    liteLoaderEndsDate  = timeFromS3Time("2017-09-18T08:39:46+00:00");
+    fabricBeginsDate    = timeFromS3Time("2019-04-23T14:52:44+00:00");
+    neoForgeBeginsDate  = timeFromS3Time("2023-09-20T09:02:57+00:00");
```
phit: earliest neoforge is 1.20.1 https://github.com/phit/multimc-meta-upstream/blob/develop/neoforge/version_manifests/1.20.1-47.1.5.json#L4

arthomnix: I already fixed that https://github.com/MultiMC/Launcher/pull/5323/commits/4fe664204e87d562172153d874985d1a3cda2c79

Issue #2692, opened 2019-06-17
Fluffy Kitty button does not meow


kaosjr:
System Information
-----------------------------
MultiMC version:
0.6.5
Operating System:
Win 10
Summary of the issue or suggestion:
----------------------------------------------
The Fluffy Kitty button should meow when you press it.

What should happen:
------------------------------
1. You press the kitty button
2. It meows and displays the image

Steps to reproduce the issue (Add more if needed):
-------------------------------------------------------------
1. Press the Kitty button

2. Notice the lack of meow

3. Be saddened

Suspected cause:
---------------------------
Noone added a meow sound yet.

Logs/Screenshots:
----------------------------
[//]: # (Please refer to https://github.com/MultiMC/MultiMC5/wiki/Log-Upload for instructions on how to attach your logs.)
![image](https://user-images.githubusercontent.com/16449077/59577818-1d178000-907a-11e9-968f-3d4640fe5dcd.png)

Additional Info:
---------------------------
If you need a meow sound, I can meow for you. It's for a good cause!

JeSuisUnCaillou:
Can confirm, I reproduced it on my computer and I was indeed saddened.

I look forward for the meow to be implemented as soon as possible, this is critical.

dolphin-cat:
This is critical we need the Fluffy Kitty button!

TheEssem:
this issue alone makes multimc literally unusable, please fix

imichaeldotorg:
What's the ETA here?

Pachin0:
Just switched to multimc, please fix so I can hear the meow

dannyhpy:
Still no ETA? :(

Hexality:
I would love if there was a meow

Would love even more if both the cat and the sound were customizable sp we can change to our own sprite/sound of something

(I'm not against the cat, I just prefer foxes over cats :3)

panda-senpai:
Confirmed the issue.  Lack of audio feedback leads to a poor experience.  Lack of customizable cat hinders usage.

devBob027:
This problem makes the launcher borderline unusable please fix ASAP

CR4KH34D8U2:
Still broken 3 years later

yeeterlol:
It's been almost 3 years. MultiMC is a horrible launcher, we need the Fluffy Kitty button to MEOW

cyphgithub:
Why is this still an issue? this needs to be resolved asap please

[...]

PR #5360, opened 2024-03-03
GH-2692 Make the cat meow

Error504TimeOut:
Now, when the "Fluffy Kitty"-Button is pressed, the cat finally meows using a sound also used in minecraft. See #2692


yeeterlol:
they murdered the cat

PR #5379, opened 2024-03-27
GH-5372 Fix FTB App import

### LINKED ISSUE #5372: FTB App "settings.json" not found
### Operating System

Windows

### Description of bug

I have FTB App version 1.25.4 for windows and encountered problem with importing instances from it - Instance list was empty.
MultiMC log-file says
`3.978 C Exception: "Unable to open C:/Users/mouse/AppData/Local/.ftba/bin/settings.json for reading: No such file or directory"
`
Seems like FTB App moved settings.json from ".ftba/bin/" to ".ftba/storage/"

### Steps to reproduce

Run MultiMC, select Add Instance and go FTB App section.
List'll be empty.

### Suspected cause

_No response_

### This issue is unique

- [X] I have searched the issue tracker and did not find an issue describing my bug.

phit:
closes GH-5372
settings path was changed with https://github.com/FTBTeam/FTB-App/commit/32c618c8fd16d794e5a7c4ac31f353b4b2ee3306

### DIFF
--- a/launcher/ui/pages/modplatform/import_ftb/Model.cpp
+++ b/launcher/ui/pages/modplatform/import_ftb/Model.cpp
@@ -93,7 +93,7 @@ QString getFTBAPath() {
 #endif
 
 QString getFTBASettingsPath() {
-    return FS::PathCombine(getFTBAPath(), "bin/settings.json");
+    return FS::PathCombine(getFTBAPath(), "storage/settings.json");
 }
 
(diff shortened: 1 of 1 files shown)


Zetabite:
I would suggest that we also look in the legacy path, people might not update the ftb app, or reuse profiles for a new instance, MultiMC not knowing the old location would force them to use the ftb app again if they really don't need to

PR #5430, opened 2024-05-31
NOISSUE Customizable Patreon URL

sasha0552:


Zetabite:
I dont see why we would ever need a customizable URL for patreon.

sasha0552:
Debranding improvement. Mostly to make life easier for those who decide to fork MultiMC.
In the current implementation, the button is even called "Support %1" (e.g. "Support DevLauncher") instead of "Support MultiMC", but the link always points to the MultiMC patreon.

Related #4718

PR #5431, opened 2024-05-31
NOISSUE Customizable news URL

sasha0552:


Zetabite:
Same as to your other PR, why would we need that?

PR #5458, opened 2024-07-06
NOISSUE: api.paste.ee => api.pastee.dev For Log Upload

TacoGS:
This changes the upload log API from the old https://api.paste.ee to https://api.pastee.dev as advised by the owner of the pastebin system.

This was tested with https://pastee.dev/p/luWOw


peterix:
Well, not doing anything ended up being the right call in the end.

PR #5481, opened 2024-07-29
NOISSUE Centralize Mojang URLs

sasha0552:
This makes it easy to change these URLs to custom Yggdrasil-compatible authentication servers, which can be used by users who do not want to use Microsoft authentication servers due to privacy issues (e.g., chat reporting feature, etc).


sylv256:
Wouldn't this allow piracy which is a liability for the project?

Zetabite:
People are pirating with or without it. It doesn't make a difference. It makes life easier for us

PR #5652, opened 2025-08-08
CMake: Require and use Java 1.8

davidebeatrici:
The specific version is mentioned in the build instructions at: https://github.com/MultiMC/Launcher/blob/abc582466ee7a88a398604c11b8ccb6be424c1fe/BUILD.md

Without this patch compilation fails because target and source have a minimum value of 8.

### DIFF
--- a/libraries/javacheck/CMakeLists.txt
+++ b/libraries/javacheck/CMakeLists.txt
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 3.1)
 project(launcher Java)
-find_package(Java 1.7 REQUIRED COMPONENTS Development)
+find_package(Java 1.8 REQUIRED COMPONENTS Development)
 
 include(UseJava)
 set(CMAKE_JAVA_JAR_ENTRY_POINT JavaCheck)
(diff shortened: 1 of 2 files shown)


davidebeatrici:
I just realized there is an identical commit in PolyMC: https://github.com/PolyMC/PolyMC/commit/8e842d1b6e695f714c8fcb43270e7545b400b68f

phit:
why would you want to drop support for java 7? theres plenty of old mods that do not work on java 8

Moresteck:
> why would you want to drop support for java 7? theres plenty of old mods that do not work on java 8

If I may ask, which mods for example?

Moresteck:
As for the PR, I agree using Java 8 for building would make the process easier, but I don't see why source and target versions should get bumped from 7 to 8. The source is still compatible with Java 7.
I'd suggest them to be left at 7 for both NewLaunch.jar and JavaCheck.jar.

davidebeatrici:
The problem is that recent OpenJDK versions support a minimum version of 8.

Moresteck:
> The problem is that recent OpenJDK versions support a minimum version of 8.

But you'd be using JDK 8 to build MultiMC, no? At least that's what I would do.

davidebeatrici:
Yes, OpenJDK 8 is still supported. If you all agree I will drop the source and target changes.

PR #5707, opened 2025-12-16
Claude/test multimc frontend x uzts

JayleKab:


Janrupf:
@JaydenKable eh? Why do you keep opening and then closing PR's? Its creating quite a bit of noise

TacoGS:
Leave AI off this project. Claude can stay where it is.

PR #5740, opened 2026-03-13
merge

Kxywek:


TacoGS:
Who would think it was a good idea to make a pull request for a pirated launcher that is a downstream fork of this was a good idea. No only did you get yourself an immediate close, you're going to be banned. Do not be like this person and try to PR a pirate launcher.

PR #5751, opened 2026-04-12
Hotfix/javadetect25

TacoGS:
This fixes the problem where the new versions of Minecraft 26 (Snapshots and beyond) are not reporting the need for Java 25 and reported Java 21 instead.


TacoGS:
Duplicate.

